PR #1, opened 2024-08-10
Add  Multithreading to fix Cathode Heating performance issues

mslaffin:
During testing on 08/09, I observed performance issues that are likely caused by synchronous serial communication with the 9104 power supplies, which block the main thread and cause substantial lag in the application. When multiple power supplies are connected simultaneously, the delays in the responses sum together and make the application very unresponsive and nearly unusable. 

This PR will involve changes to implement asynchronous I/O operations allow the program to continue executing while waiting for I/O operations to complete. 

Changes at this time should be limited to the 9104PowerSupply class, and updates to CathodeHeatingSubsystem to accommodate the modified instrumentctl driver.

### DIFF
--- a/instrumentctl/power_supply_9104.py
+++ b/instrumentctl/power_supply_9104.py
@@ -3,6 +3,7 @@ import threading
 import time
 from utils import LogLevel
 import os
+import queue
 
 class PowerSupply9104:
     MAX_RETRIES = 3 # 9104 display display reading attempts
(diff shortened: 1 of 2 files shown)


mslaffin:
I've spent some time implementing a Producer-Consumer pattern to offload the 9104 communication from the main application thread. 

Changes to `PowerSupply9104` class thus far:

- Added a `command_queue` method to `PowerSupply9104` to act as a buffer between the producer (main thread) and the consumer (worker thread)
- The main thread "produces" commands by calling `enqueue_command`, which adds communication tasks to the queue for each supply. 
- Each `PowerSupply9104` instance has its own `_command_worker` thread which "consumes" these commands, and executes them in the order they were received
- The goal here is to fix the lag issue by allowing the main thread to quickly delegate tasks without slowing down GUI updates while waiting for their completion. 

Changes to `CathodeHeatingSubsystem` class:
- The `initialize_power_supplies` method now implements a chain of callbacks to handle the initialization sequence 
    - previously `initialize_power_supplies` waited for the entire setup process to complete before moving to the next PS
    - Each step is now separated into callback methods: set preset mode -> confirm preset mode -> set OVP -> confirm OVP -> set OCP -> confirm OCP
    - The callback methods are passed to `enqueue_command` and are executed when the command completes
 - The `update_data` method now acts more as a scheduler, initiating update request at the defined frequency rather than performing the updates directly

PR #5, opened 2024-09-11
COM port update handling

mslaffin:
Updates to the COM port functionality and initialization processes for CCS and VTRX.

Adding `update_com_port` methods to each subsystem allow changing the port configuration during runtime. Serial connection management, handling, etc.


mslaffin:
These changes are stale and Mark/Arayan have come up with a better solution so I'm going to close this without merging.

PR #8, opened 2024-10-01
Avoid overloading vtrx queue with None. Message rate limiting

mslaffin:
From asana: [https://app.asana.com/0/1208196184307227/1208357297655993](https://app.asana.com/0/1208196184307227/1208357297655993)

Testing with hardware on 2024-09-17 revealed an issue with the VTRX system communication. The VTRX system was powered on prior to running the dashboard application. The  GUI hung when trying to initialize serial connection, and then printed this message repeatedly: 
```
[17:57:22] - WARNING: Incomplete data received.
```
[log_2024-09-17_17-58-51 (1).txt](https://github.com/user-attachments/files/17198439/log_2024-09-17_17-58-51.1.txt)

### DIFF
--- a/subsystem/vtrx.py
+++ b/subsystem/vtrx.py
@@ -58,6 +58,9 @@ class VTRXSubsystem:
             1: tk.PhotoImage(file=resource_path("media/on.png"))
         }
         self.error_state = False
+        self.last_incomplete_data_warning_time = 0
+        self.last_data_received_time = time.time()
+        self.warning_interval = 5  # rate limit to once per 5 seconds
         self.setup_serial()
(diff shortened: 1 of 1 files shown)


mslaffin:
I believe application was being put into a bad state and hanging due to the `read_serial` method continuously adding `None` to `data_queue` when the port was connected but data wasn't received properly. If this was the case, the process_queue method repeatedly calling `update_gui_with_error_state`, might've overload the main GUI thread and caused it to be sluggish. I've bypassed this, and limited the rate at which error messages are logged. This should be tested again to be sure it is fixed

PR #13, opened 2024-10-11
Bugfix/E5CN-rework

mslaffin:


mslaffin:
After updating pymodbus from `3.6.9` to `3.7.3`, I was able to make several changes successfully read process registers during 10/21 testing with the expected format:

image


usage from the python interpreter:
```
>>> from instrumentctl.E5CN_modbus import E5CNModbus
>>> e5cn = E5CNModbus(port='COM16', baudrate=9600, parity='E', stopbits=2, bytesize=8)
DEBUG: Initializing E5CNModbus with port: COM16
>>> e5cn.connect()
True
>>> e5cn.read_temperature(unit=1)
DEBUG: Received registers: [0, 203]
INFO: Temperature from unit 1: 20.30 °C
20.3
```

Sample output 
test_with_real_tempdata

Several problems need to be solved here relating to CathodeHeatingSubsystem:
- Initiating dashboard with one connected controller and two disconnected controllers introduced massive lag. Is blocking main event loop
- System can not re-establish USB connection after disconnect
- Have not tested with all three controllers connected
- Have not tested with solid thermocouple connection
- Have not tested memory consumption over time

mslaffin:
I've confirmed that the core E5CN functionality is working with the CCS hardware as of 11/04. I was able to read temperature continuously from all three controllers on the same bus, with no lag or disruption to the GUI.  

During this testing, several other problems cropped up, but I believe they are unrelated to these changes, so I will be merging this. See attached screenshot and log file:
[Uploading log_2024-11-04_17-48-06.txt…]()
Capture_three_E5CN_custom_harness

bwalkerMIR:
This is awesome!

PR #15, opened 2024-10-13
Feature/9104 voltage ramping implementation

mslaffin:
This feature introduces voltage ramping functionality for the cathode heating subsystem.

- `ramp_voltage` now accepts `step_size` and `step_delay` parameters
- ramp operation occurs in separate `_ramp_voltage_thread`
- set voltage is zeroed when enabling output

### DIFF
--- a/instrumentctl/power_supply_9104.py
+++ b/instrumentctl/power_supply_9104.py
@@ -113,7 +113,7 @@ class PowerSupply9104:
             self.log(f"Error setting current: {error_message}", LogLevel.ERROR)
             return False
 
-    def ramp_voltage(self, target_voltage, ramp_rate=0.01, callback=None):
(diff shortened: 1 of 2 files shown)


mslaffin:
Tested this on 10/16. [Link ](https://drive.google.com/file/d/1xVLMUyPTTjNm4VCtXEgVdFgmyhOmQPd-/view?usp=sharing)to google drive video

The reported least significant value has noise, but overall the ramp is linear. I took this data with the supply output unloaded, programmed with the default ramp rate of 0.01V/s. The attached log and plot show it ramping from 0-0.5V, then disabling for ~30 seconds, followed by another ramp to 0.75V 
Voltage ramping
[log_2024-10-16_16-13-59_ramped_to_0p75_at_0p01VperSec.txt](https://github.com/user-attachments/files/17423959/log_2024-10-16_16-13-59_ramped_to_0p75_at_0p01VperSec.txt)

PR #16, opened 2024-10-14
Feature/G9_Driver

mslaffin:
Opening this for discussion of initial work on G9 driver code.


instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,254 @@
...
+    def send_command(self):
+        if not self.is_connected():
+            raise ConnectionError("Seiral Port is Not Open.")
+        header = b'\x40\x00\x00\x0F\x4B\x03\x4D\x00\x01' 
+        data = b'\x00\x00\x00\x00\x00\x00'
+        self.msgOptData = data
+        checksum = self.calculate_checksum(header + data, 0 , len(header + data))
+        cs = b'\x00' + checksum if len(checksum) == 1 else checksum
+        footer = b'\x2A\x0D' 
+        self.ser(header + data + cs + footer)
```
mslaffin, line 52: I don't think `self.ser` can be called with this data. Probably has to be `self.ser.write(header + data + cs + footer)` instead.

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,254 @@
...
+                    raise ValueError("Error was detected in inputs but was not found")
+                
+            SOTSF = data[27:31]
+            if not self.check_flags13(SOTSF):
+                if self.safety_out_terminal_error(data[55:71]):
+                    raise ValueError("Error was detected in outputs but was not found")
+                
+            # Unit status
+            US = data[73:75]
+            if US != 0:
```
mslaffin, line 127: looks like this attempts to compare a bytes object with an integer. 

should probably be something like `int.from_bytes(US)`

scripts/G9Drivertest/g9_driver.py
```diff
@@ -0,0 +1,254 @@
...
+                gates = self.bytes_to_binary(SITDF)
+                for i in range(18):
+                    if gates[-i + 1] == "0":
+                        err.append(i)
+                raise ValueError("An input is either off or throwing an error: {err}")
+            
+            SOTDF = data[17:21]
+            if not self.check_flags13(SOTDF):
+                err = []
+                gates = self.bytes_to_binary(SITDF)
```
mslaffin, line 108: This should be `SOTDF` not `SITDF`

mslaffin:
I think `response` needs additional consideration. We should be parsing this according to the specified format
image

`response` currently assumes a fixed response length of 199 bytes which seems fragile to me. We could instead parse out the actual response length specified in the response header (represented in bytes +1 to +3) to know how much data to expect. We could read out the header first and check that it's complete (4 bytes)
```python
header = self.ser.read(4)
    if len(header) != 4:
        raise ValueError("Incomplete response header.")
    if header[0] != 0x40:
        raise ValueError("Invalid response start code.")
```
then calculate the response length and raise errors if it's invalid
```python
response_length = int.from_bytes(header[1:4], byteorder='big')
if response_length <= 0:
    raise ValueError("Invalid response length.")
```
then read the rest of the response data
```python
    remaining_length = response_length + 4  # +4 for checksum and footer
    data = self.ser.read(remaining_length)
    if len(data) != remaining_length:
        raise ValueError("Incomplete response data.")
```

then continue to parse out specific codes, checksums, etc. as closely as possible to the specified protocol

mark11778:
Ok so i have added to check the length of the message and to process the header code. But correct me if I am wrong, if the header code is not 0x40, and is either and error(0x09) or the incorrect format(0x06) there is no need to check the rest of the message data because it is basically empty, right?

The incorrect format ending at the service code section and the error format only containing 2 bytes of reserve data in the data section.

So we really don't have to do anything when it has the first byte that isn't 0x40, outside of raising an error.

mark11778:
Also a side note but from testing with the G9 being set up in it's circuit the error flags are not doing what they are suppose to ... my understanding. I could be wrong the driver I think is throwing the right error, refer to the test.txt contents to see the responses. I can get some photos of the lights on the g9 so show what I am trying to say

mslaffin:
> Ok so i have added to check the length of the message and to process the header code. But correct me if I am wrong, if the header code is not 0x40, and is either and error(0x09) or the incorrect format(0x06) there is no need to check the rest of the message data because it is basically empty, right?

I agree that we shouldn't have to process the rest of the message data if the first byte `+0` is not `0x40`. If `+0` is `0x40` and `+3` is `0x09` or `0x06`, then we also don't want to process the rest of the message, but we should handle and log the specific error based on the End Code. 

If `+0` is `0x40` and `+3` is `0xC3`, only then we should proceed to process the full response packet.

mslaffin:
I'm noticing there's two duplicate G9 driver files in this branch. 
- EBEAM_dashboard/instrumentctl/g9_driver.py
- EBEAM_dashboard/scripts/G9Drivertest/g9_driver.py

This is confusing in a collaborative project. For instance, I left [this comment](https://github.com/uw-loci/EBEAM_dashboard/pull/16#discussion_r1798636051) yesterday which @mark11778 resolved in the G9Drivertest/g9_driver.py, but not in the canonical instrumentctl/g9_driver.py. Let's try and avoid committing duplicate files to the repo. I'm all for creating spin-off scripts for test purposes, but keeping this feature branch clean from duplicated files (also usr config data, compiled bytecode files, virtual environments, etc.) will help everyone out.

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,264 @@
...
+        self.ser.write(header + data + reserve+ cs + footer)
+
+        self.response()
+
+    # used mainly for the check sum but can also be used to check for error flags
+    # needs an input of a byte string and the range of bytes that need to be sum
+    # will return the sum of the bytes in the a byte string in the form of b'\x12'
+    def calculate_checksum(self, byteString, startByte, endByte):
+        assert isinstance(byteString, bytes)
+        return sum(byteString[startByte:endByte + 1]).to_bytes(1, "big")
```
mslaffin, line 64: guys I think I see an issue with this checksum method. Has anyone confirmed if this actually works as expected?

Pg. 116 of the datasheet specifies that the checksum is 16 bits (2 bytes) long. The `to_bytes` built-in in this line explicitly sets the length to **1 byte**:
```python
return sum(byteString[startByte:endByte + 1]).to_bytes(1, "big")
```

I think this would truncate the checksum to the least significant 8 bits, and discard any higher bits?

bwalkerMIR:
> Also a side note but from testing with the G9 being set up in it's circuit the error flags are not doing what they are suppose to ... my understanding. I could be wrong the driver I think is throwing the right error, refer to the test.txt contents to see the responses. I can get some photos of the lights on the g9 so show what I am trying to say

Can you explain this more? What is not working right?

bwalkerMIR:
> I'm noticing there's two duplicate G9 driver files in this branch.
> 
> * EBEAM_dashboard/instrumentctl/g9_driver.py
> * EBEAM_dashboard/scripts/G9Drivertest/g9_driver.py
> 
> This is confusing in a collaborative project. For instance, I left [this comment](https://github.com/uw-loci/EBEAM_dashboard/pull/16#discussion_r1798636051) yesterday which @mark11778 resolved in the G9Drivertest/g9_driver.py, but not in the canonical instrumentctl/g9_driver.py. Let's try and avoid committing duplicate files to the repo. I'm all for creating spin-off scripts for test purposes, but keeping this feature branch clean from duplicated files (also usr config data, compiled bytecode files, virtual environments, etc.) will help everyone out.

I agree.

instrumentctl/g9_driver.py
bwalkerMIR: > guys I think I see an issue with this checksum method. Has anyone confirmed if this actually works as expected?
> 
> Pg. 116 of the datasheet specifies that the checksum is 16 bits (2 bytes) long. The `to_bytes` built-in on this line explicitly sets the length to **1 byte**:
> 
> ```python
> return sum(byteString[startByte:endByte + 1]).to_bytes(1, "big")
> ```
> 
> I think this would truncate the checksum to the least significant 8 bits, and discard any higher bits?

They did get the checksum to work with the test code which was last week which I assume is either similar or the same. The G9 has 2 bytes for the checksum, but the checksum with the zero optional data we send is always fully contained within the low byte, which is likely why they throw away the high byte. This may be poor practice though, for example, if we decided to send optional data at some point I haven't done the math to determine if we could end up with non zero high byte data. I think we should also calculate the full check sum and send the high byte and the low byte, which in our case most of the time (possibly all) the high byte will be zero.

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,264 @@
...
+        else:
+            return False
+
+    def response(self):
+        if not self.is_connected():
+            raise ConnectionError("Seiral Port is Not Open.")
+        
+        data = self.ser.read_until('b\r')
+        self.lastResponse = data
+        if data[0] == b'@':
```
mslaffin, line 89: I think this `if data[0] == b'@'` comparison is incorrect. Indexing a bytes object like data[0] should return an integer and not another bytes object. we should instead check if `if data[0] == 0x40:`

for reference:
```
>>> test_bytes = b'TEST'
>>> test_bytes[0]
84
>>> type(test_bytes[0])
<class 'int'>
>>> bytes([test_bytes[0]])
b'T'
>>>
```

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,264 @@
...
+    # of a byte string
+    def check_flags13(self, byteString, norm = 1):
+        assert isinstance(byteString, bytes)
+        # this is for if we only need the last 13 bits (more or less hardcoding this 
+        # just including the rest if it might be helpful in the future
+        if int(self.bytes_to_binary(byteString)[-13:], 2) >= (13 * norm):
+            # all flags we care about are 1
+            return True
+        else:
+            return False
```
mslaffin, line 81: My understanding is that this method might be used to check if the status bits in the given bytes object are all set to 1. but it looks like there's a type conversion problem here that might cause problems.  

The condition `if int(bits[-13:], 2) >= (13 * norm):` only checks if the integer value of the bitstring >= 13, not that all the status flags are 1's. The check will incorrectly return True on some bitstrings that contain 0's (error flags)

converting 13 bits into an integer you get 2^13 - 1= 8191
```
>>>
>>> def check_flags13(byteString, norm=1):
...     assert isinstance(byteString, bytes)
...     binary_string = bytes_to_binary(byteString)
...     last_13_bits = binary_string[-13:]
...     integer_value = int(last_13_bits, 2)
...     if integer_value >= (13 * norm):
...         return True
...     else:
...         return False
...
>>> def bytes_to_binary(byte_string):
...     return ''.join(format(byte, '08b') for byte in byte_string)
...
>>> byteString = bytes([0xFF, 0xFF])
>>> print(bytes_to_binary(byteString))
1111111111111111
>>> check_flags13(byteString)
True
>>> errorByteString = bytes([0x00, 0x1F])
>>> print(bytes_to_binary(errorByteString))
0000000000011111
>>> check_flags13(errorByteString)
True
>>>
```

instrumentctl/g9_driver.py
mslaffin: okay yea that makes sense. In any case I think truncating the checksum undermines the purpose of the checksum. and we shouldn't rely on an assumption that the high byte will always be zero. esp because this is part of a safety system

instrumentctl/g9_driver.py
bwalkerMIR: I agree it seems like checking >=13 is not the same as all flags being 1

arundhatisingh17: We're comparing the string value of the binarystring to that of 13 1's now.

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,274 @@
...
+    def response(self):
+        if not self.is_connected():
+            raise ConnectionError("Serial Port is Not Open.")
+        
+        data = self.ser.read_until('b\r')
+        self.lastResponse = data
+
+
+        # Indexing such that we don't return an integer
+        if data[0:1] == '\x40':
```
AryanSV5, line 99: Slicing the data works and I also swapped the character symbols to the hexadecimal byte representations. I tested this via an online python interpreter.

instrumentctl/g9_driver.py
mark11778: Yeah I wrote that, and tested it. I can do a little more follow up testing to make sure it works in all cases( including with optimal data). I will document these tests somewhere.

instrumentctl/g9_driver.py
mark11778: In the int casing this coverts the int from binary to decimal with the 2 argument. There was two reasons I went with this method, one being that the flags are not the same for all inputs for some 1 means error and others 0 means error. Thus this method could be used for both given the norm argument. Along with that it also gives an easier way to change the number of inputs. But if we think a different way is better, we can go with that.

Screenshot 2024-10-17 081223

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,273 @@
...
+        #     # all flags we care about are 1
+        #     return True
+        # else:
+        #     return False
+
+    def response(self):
+        if not self.is_connected():
+            raise ConnectionError("Serial Port is Not Open.")
+        
+        data = self.ser.read_until('b\r')
```
mslaffin, line 94: This terminator is incorrect. It's a string, not a bytes object and won't match on the end sequence properly.
```
>>> type('b\r')
<class 'str'>
>>> type(b'\r')
<class 'bytes'>
```
This might've gone unnoticed because `read_until` [inherits the 0.5 second port timeout](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.read_until), and returns whatever it read. The problem I foresee is this blocking the main event loop every time it waits for the timeout. I ran into a similar problem with the 9104 power supplies.
Per the response format, the actual terminator sequence is `b'\x2A\x0D'` or `b'*\r'`

arundhatisingh17: fixed.

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+FOOTER = b'\x2A\x0D'
+
+class G9Driver:
+    def __init__(self, port=None, baudrate=9600, timeout=0.5, logger=None, debug_mode=False):
+        # not sure if we will need all of these but just adding them now in case
+        self.US, self.SITDF, self.SITSF, self.SOTDF, self.SOTSF, self.lastResponse = None, None, None, None, None, None
+        if port:
+            self.ser = serial.Serial(port, baudrate, parity=serial.PARITY_EVEN,
+                                      stopbits=serial.STOPBITS_ONE, bytesize=serial.EIGHTBITS,
+                                        timeout=timeout)
```
mslaffin, line 51: we should ensure `self.ser` is defined even if `port` is not provided. 
```
if port:
    self.ser = serial.Serial(
	port=port,
	baudrate=baudrate,
	parity=serial.PARITY_EVEN,
	stopbits=serial.STOPBITS_ONE,
	bytesize=serial.EIGHTBITS,
	timeout=timeout
    )
else:
    self.ser = None
```

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+        assert isinstance(byteString, bytes)
+        binary_string = self.bytes_to_binary(byteString)[-NUMIN:]
+        string_of_ones = norm * NUMIN
+        return binary_string == string_of_ones
+    
+    def response(self):
+        if not self.is_connected():
+            raise ConnectionError("Serial Port is Not Open.")
+        
+        data = self.ser.read_until(b'\r')
```
mslaffin, line 91: This still doesn't follow the G9 manual:
image

I can't think of a good reason not to read until the full footer is received. should be:
```
data = self.ser.read_until(FOOTER)
```

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+        self.lastResponse = data
+
+        # Indexing such that we don't return an integer
+        if data[0:1] == b'\x40':
+            if data[3:4] == b'\xc3':
```
mslaffin, line 96: instead of comparing slices, why can't we just access these bytes directly?
```
if data[0] == 0x40:
    if data[3] == 0xC3:
```

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+    1: Invalid configuration
+    2: External test signal failure
+    3: Internal circuit error
+    4: Discrepancy error
+    5: Failure of the associated dual-channel input
+    """
+    # checks all the SITSFs, throws error is one is found
+    def safety_in_terminal_error(self, data):
+        if len(data) != 10:
+            raise ValueError(f"Expected 24 bytes, but received {len(data)}.")
```
mslaffin, line 169: confusing message here. `len(data) != 10` but then says `Expected 24 bytes`. Shouldn't this match the actual expected lengths?

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+    
+    # this method is made to check the error flags, right not only checks the last 13 bits
+    # of a byte string
+    def check_flags13(self, byteString, norm = '1'):
+        assert isinstance(byteString, bytes)
+        binary_string = self.bytes_to_binary(byteString)[-NUMIN:]
+        string_of_ones = norm * NUMIN
+        return binary_string == string_of_ones
+    
+    def response(self):
```
mslaffin, line 87: it looks like we're still not using `calculate_checksum` in the response method. 

I think we still need to add code here to calculate the checksum of the received data and compare it with the checksum provided in the message itself

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+                if alwaysHeader != RECHEADER or alwaysFooter != FOOTER:
+                    raise ValueError("Always bits are incorrect")
+                
+                # Save all the msg data so backend can access before checking for errors
+                self.US = data[73:75]
+                self.SITDF = data[11:17]
+                self.SITSF = data[21:27]
+                self.SOTDF = data[17:21]
+                self.SOTSF = data[27:31]
+
```
mslaffin, line 108: All of these data offsets should be defined as constants
```
US_OFFSET = 73
SITDF_OFFSET = 11
SOTDF_OFFSET = 17
SITSF_OFFSET = 21
SOTSF_OFFSET = 27
``` 
then used like
```
self.US = data[US_OFFSET:US_OFFSET + 2]
self.SITDF = data[SITDF_OFFSET:SITDF_OFFSET + 6]
```

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+    # this function will need to be constantly sending requests/receiving to check when the high voltage is off/on
+    def checkStatus():
+        pass
+
+    def flush_serial(self):
+        self.ser.reset_input_buffer()
+
+    # this just makes sure that the ser object is considered to be valid
+    def is_connected(self):
+        return self.ser.is_open
```
mslaffin, line 251: probably should be `return self.ser is not None and self.ser.is_open`

instrumentctl/g9_driver.py
```diff
@@ -0,0 +1,255 @@
...
+
+NUMIN = 13
+
+SNDHEADER = b'\x40\x00\x00\x0F\x4B\x03\x4D\x00\x01' 
+SNDDATA = b'\x00\x00\x00\x00'
+SNDRES = b'\x00\x00'
+
+RECHEADER = b'\x40\x00\x00'
+
+FOOTER = b'\x2A\x0D'
```
mslaffin, line 42: these constants are specific to the G9Driver and won't be used or imported elsewhere, so I'd like these to be encapsulated within the class. Also these should be in all caps like:
```
class G9Driver:
    IN_STATUS = {
        0: "No error",
        1: "Invalid configuration",
        2: 'External test signal failure',
        3: 'Internal circuit error',
        4: 'Discrepancy error',
        5: 'Failure of the associated dual-channel input'
    }
```

instrumentctl/g9_driver.py
arundhatisingh17: changed

instrumentctl/g9_driver.py
arundhatisingh17: changed

instrumentctl/g9_driver.py
AryanSV5: Good call, just added it.

instrumentctl/g9_driver.py
AryanSV5: That is a typo. The 'Safety Input Terminal Error Causes' section of the data is 24 bytes but only 10 bytes indicate the error causes whilst the remaining are reserved. We accidentally put in 24 instead of 10.

mslaffin:
Looks like `calculate_checksum` included an extra end byte that may have caused an off-by-one problem. I had all tests passing after stripping off the old invalid checksums from the saved msg data and re-adding it using the fixed method.
```
(venv) PS C:\Users\Admin\EBEAM_dashboard> python -m unittest .\instrumentctl\errorTests.py -v
test_checksum_validation (instrumentctl.errorTests.TestG9Driver.test_checksum_validation) ... ok
test_input_error (instrumentctl.errorTests.TestG9Driver.test_input_error) ... ok
test_invalid_length (instrumentctl.errorTests.TestG9Driver.test_invalid_length) ... ok
test_invalid_start_byte (instrumentctl.errorTests.TestG9Driver.test_invalid_start_byte) ... ok
test_no_error (instrumentctl.errorTests.TestG9Driver.test_no_error) ... ok
test_output_error (instrumentctl.errorTests.TestG9Driver.test_output_error) ... ok
test_unit_state_error (instrumentctl.errorTests.TestG9Driver.test_unit_state_error) ... ok

----------------------------------------------------------------------
Ran 7 tests in 0.010s

OK
```

subsystem/interlocks.py
```diff
@@ -1,60 +1,157 @@
...
+    12 : "G9SP Active" # G9SP Active
+    }
+
+def handle_errors(self, data):
+    try:
+        response = g9_driv.read_response()
+        return {"status":"passes", "message":"No errors thrown at this time."}
+
+    except ValueError as e:
+        return {"status":"error", "message":str(e)}
```
mslaffin, line 31: This is being defined outside the class (`self` makes no sense as a param) and not used. I think this should be deleted and implement error handling directly in `update_data`

subsystem/interlocks.py
```diff
@@ -1,60 +1,157 @@
...
+        for i, (k,v) in enumerate(self.indicators.items()):
+            tk.Label(interlocks_frame, text=f"{k}", anchor="center").grid(row=0, column=i*2, sticky='ew')
+            canvas, oval_id = create_indicator_circle(interlocks_frame, 'red')
+            canvas.grid(row=0, column=i*2+1, sticky='nsew')
+            self.indicators[k] = (canvas, oval_id)
+
+    # logging the history of updates
+    def update_interlock(self, name, safety, data):
+        # means good
+        if safety & data == 1:
```
mslaffin, line 100: I think there's an issue with this logic here. The `==` operator has higher precedence than `&` so this is evaluated as `safety & (data == 1)` not `(safety & data) == 1` (which is how I think it should be)

mslaffin:
hey guys, I've made some major changes to to the structure of how we do status and error checking. Please take a look and let me know if you have specific questions or see any issues. 

also I created an updated test suite that focuses on testing the G9 driver's response parsing and error handling capabilities. Please review this for accuracy. (you should be able to import the class from instrumentctl now as well)
```
(venv) PS C:\Users\Admin\EBEAM_dashboard\scripts\G9Drivertest> python -m unittest errorTests.py
----------------------------------------------------------------------
Ran 9 tests in 0.003s

OK
```

bwalkerMIR (commented):
This "overhaul G9 data access and comm handling" commit is a good change, improving error handling, readability, and code maintainability. I agree with the changes and coding improvements in this commit. In the future there should be a readme next to subsystems showing if there are any tests that exist on current code so that the person commiting code is aware of any failing tests from code updates. With a big change like this more communication explaining the big changes would be helpful for all to reduce surprises, although it shouldn't come at the expense of slow down code development and building good code fast. A good medium ground would probably be coding it up and commiting it for all to review, but briefly making a twist post about the changes to summarize what the main big changes are simultaneously with committing.

bwalkerMIR:
The brief explanation could also just go here it doesn't have to go in twist. Github really is great!

instrumentctl/g9_driver.py
```diff
@@ -107,255 +172,74 @@ def calculate_checksum(self, data, start=0, end=194):
...
-            
-        if bits[-1] == "0":
-            raise ValueError(f"Unit State Error: Normal Operation Error Flag (bit 0)")
-
     #TODO: Check to see if the G9 switch is allowing high Voltage or not
     # this function will need to be constantly sending requests/receiving to check when the high voltage is off/on
     def checkStatus():
         pass
 
-    def flush_serial(self):
```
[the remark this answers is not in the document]
mslaffin: Is this comment referring to the deletion of `flush_serial`? My rationale for removing it was that `flush_serial` was defined here, but looked like it was never actually called or used anywhere. I also think this was pulled from the power_supply_9014 driver where I had originally used for debugging purposes but is not necessary.

instrumentctl/g9_driver.py
```diff
@@ -62,42 +54,115 @@ class G9Driver:
...
+                    ) 
+                self.log(f"Serial connection established on {port}", LogLevel.INFO)
+            except serial.SerialException as e:
+                self.ser = None
+                self.log(f"Failed to open serial port {port}: {str(e)}", LogLevel.ERROR)
+        else:  
+            self.ser = None
+            self.log("No port specified", LogLevel.WARNING)
+
+    def get_interlock_status(self):
```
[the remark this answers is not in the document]
mslaffin: The primary issue I saw with the earlier version of `send_command` was that it didn't actually return any status data when called in InterlocksSubsystem. The data received was stored as G9Driver class attributes rather than being returned to the caller. 

My understanding is that this violated OOP design conventions, and is a poor design because the caller has no way of knowing if the command actually succeeded. Storing the data as G9 class attributes and then accessing them directly from InterlocksSubsystem was also problematic because we'd have no way of knowing if and when the data becomes stale or when it was last updated.

instrumentctl/g9_driver.py
```diff
@@ -62,42 +54,115 @@ class G9Driver:
...
-
-        self.debug_mode = debug_mode
         self.logger = logger
+        self.debug_mode = debug_mode
+        self._setup_serial(port, baudrate, timeout)
+        self.last_data = None
         self.input_flags = []
 
-    def send_command(self):
+    def _setup_serial(self, port, baudrate, timeout):
```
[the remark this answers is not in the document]
mslaffin: The original serial connection setup was performed in the init method, which was not inherently wrong from my understanding, but can't be called again without recreating the whole object, and mixed the initialization of status variables with connection setup. 

I added some additional error handling here as well. Also an explicit setup_serial follows conventions I've tried to set in other instrumentctl drivers (such as the setup_serial in [PowerSupply9014](https://github.com/uw-loci/EBEAM_dashboard/blob/cae042495e4c42782b434b0e5537661789a92bf8/instrumentctl/power_supply_9104.py#L18))

subsystem/interlocks.py
```diff
@@ -1,60 +1,251 @@
...
+            self.log("Invalid inputs to _set_all_indicators", LogLevel.ERROR)
+
+        if self.INDICATORS:
+            for name in self.INDICATORS:
+                canvas, oval_id = self.INDICATORS[name]
+                current_color = canvas.itemcget(oval_id, 'fill')
+                if current_color != color:
+                    canvas.itemconfig(oval_id, fill=color)
+                    self.log(f"Interlock {name}: {current_color} -> {color}", LogLevel.INFO)
+
```
arundhatisingh17, line 180: In the update_data() method, shouldn't "if current_time - self.last_error_time > (self.update_interval / 1000):" check "if current_time - self.last_error_time > self.update_interval:" instead, given that current time and the timestamp of the last error are recorded in milliseconds as well..

mslaffin: Hmm.. I think that this is OK the way it is currently.

[time.time()](https://docs.python.org/3/library/time.html#time.time) returns time in seconds (not ms), but `self.update_interval` is stored in milliseconds (see where it's initialized to 500 ms in the init method. So to compare these properly, we need to convert `update_interval` from milliseconds to seconds by dividing by 1000

mslaffin:
@mark11778 pointed out yesterday that when connected to the G9 controller, the application hangs shortly after opening. I was able to confirm this also occurs in my own environment without connection to the actual controller. I have a theory about why this might be happening:

Every 500ms, the `update_data` method calls `get_interlock_status()`

Inside this method we call `_read_response` which famously uses `self.ser.read_until(self.FOOTER)`

if we're not receiving or interpreting the FOOTER byte sequence correctly, pyserial [read_until](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.read_until) blocks until the timeout is met. 

The timeout for the `G9Driver` pyserial object is 0.5 seconds. So every 500ms we're calling a function that likely blocks for 500ms. 

I think this could be hinting there's still problems are how it's receiving/interpreting the G9 output, and that we'll need to capture the specific response in isolation from the rest of the app to figure out whats going wrong.

Id run the following commands when connected to the setup:
```
python
```

```python
>>> from instrumentctl.g9_driver import G9Driver
>>> g9 = G9Driver(port='COM3', baudrate=9600, timeout=0.5, debug_mode=True)
>>>
>>> g9._send_command()
>>> response = g9._read_response()
>>> print(response)
>>>
>>> status = g9.get_interlock_status()
>>>
>>> print(status)

```

mslaffin:
I think we'll also want to move the serial communication to a separate worker thread so that the G9 communication _can't_ block the gui, but that should be a lower priority at this point.

mark11778:
Unit status no longer raises a ValueError, it will now only output to the log on the GUI. This change was necessary because the US flag was being checked before returning the SITSF and SITDF flags. If an error occurred, the Unit Normal Operation Flag (bit 0) would indicate the error and prevent the other flags from being returned to the GUI.

Along with this change, the way that the status flags were returned was incorrect. This was because the bytes were being received as big-endian, when they are actually stored in little-endian, with the bits inside the bytes being big-endian.

mslaffin:
Hi all, I've just added some larger changes to the driver, including threaded serial communication. 

Additionally, I changed some of the error handling for InterlocksSubsystem. 
- exponential backoff for error retries now has a capped update interval
- removed ConnectionError handling in favor of a unified exception path

Please take a moment to review these changes to see if you agree. Also, I need to recruit everyone's help in verifying this didn't break anything with the hardware. 

@AryanSV5, did you happen to save off any of those logs from your testing on Monday?

mslaffin:
Here's the log file Mark recorded during our meeting on 11/15: [log_2024-11-15_16-43-38.txt](https://github.com/user-attachments/files/17782299/log_2024-11-15_16-43-38.txt)

This test showed a new problem, where the indicators were frequently flashing red between successful reads. the log shows "No data" errors approximately every 1 second between successful responses:
```
[16:43:18] - CRITICAL: No data available from G9
[16:43:19] - CRITICAL: No data available from G9
[16:43:20] - CRITICAL: No data available from G9
```

Recent changes added a background thread with a message queue, whereas the previous version waited for a complete cycle of send/receive before returning. 

Essentially I think it is reading the queue faster than complete responses can be processed. When [_response_queue.get_nowait()](https://github.com/uw-loci/EBEAM_dashboard/blob/3ba4b0a1c8f3735411bcacc24e1b79777cbbb8a5/instrumentctl/g9_driver.py#L126) on line 126 finds no data, it immediately returns None and triggers red indicators. The G9 datasheet specifies a max receive processing of 300ms, so I think this makes sense given the frequency observed in the log. 

I am thinking that this means we should remove the step of placing None into the queue on read errors, and rely on the serial timeout to catch incomplete or missing responses.

mark11778:
There was two lines in the threading feature that led it to be problematic. Line 133 where we were calling get_nowait(), which removes the last response from the q. That along with get_interlock_status was being called faster than it could process a response, thus leading to returning None. I also think the break on line 116 was problematic, but could be possibly changed back now after the other change.

PR #19, opened 2024-10-20
Feature/manual voltage entry mod

mslaffin:
This pull request improves the UI element for manually entering a voltage value for the Cathode Heating Subsystem. 

Previously, double-clicking on the voltage label for a specific cathode revealed a pop-up window to let users enter a float (representing the 9104 power supply upper voltage limit):

HeaterV_pop-up

This has been sort of cumbersome, and requires additional keystrokes from the user. Instead with these changes the user entry field is embedded directly into the frame. This needs to be tested on CCS hardware prior to merging:
image
image


mslaffin:
Closing this as it is stale and a low priority/cosmetic improvement. Can be revisited at a later point.

PR #22, opened 2024-11-06
fix toggle_output set voltage sequencing

mslaffin:
Modified `toggle_output` in CathodeHeatingSubsystem:

- Add explicit voltage zeroing with verification before enabling output
- Set "zero" voltage tolerance to 0.07V

### DIFF
--- a/instrumentctl/power_supply_9104.py
+++ b/instrumentctl/power_supply_9104.py
@@ -131,20 +131,36 @@ class PowerSupply9104:
         thread.start()
 
     def _ramp_voltage_thread(self, target_voltage, step_size, step_delay, preset, callback):
-        current_voltage = 0.0  # Starting voltage, adjust if you can fetch from the PSU.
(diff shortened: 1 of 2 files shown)


mslaffin:
This appears to be working again. Interestingly, I found that querying the supply too rapidly can cause major communication issues, so I had to strip out some additional verification steps to reduce the interaction occuring between the `update_data` interval.

Here's a plot from the attached log that showcases the slope. Note that in this plot, the frequent drops to 0V are an artifact of the post processor code, and not reflective of the actual behavior. aka, my plotting code is still under construction and is getting caught up on some formatting. 

log_2024-11-11_17-51-51_three_ramped_simultaneously_with_temp_voltage


[log_2024-11-11_17-51-51_three_ramped_simultaneously_with_temp.txt](https://github.com/user-attachments/files/17709278/log_2024-11-11_17-51-51_three_ramped_simultaneously_with_temp.txt)

PR #23, opened 2024-11-08
consolidate color log

mslaffin:
`set_plot_alert` is called during temperature reading but is later overridden by `update_plot`. I think this has caused funky behavior in previous tests.

This PR creates a single `set_plot_color` that handles all cases, and indicates communication errors in orange as well.
Needs to be tested prior to merging


mslaffin:
I confirmed this to be working on 11/11:
Normal operation:
blue
COM disconnected:
orange
Overtemp condition on Cathode C:
overtemp

PR #25, opened 2024-11-16
E5CN Connection Instability Fix

mslaffin:
On 11/14 Chase Ott noticed there were numerous errors in the log when testing the E5CN connection on `develop`. 

Here's the log capture: 
[log_2024-11-14_15-50-15.txt](https://github.com/user-attachments/files/17787091/log_2024-11-14_15-50-15.txt)

socket disconnections and reconnection attempting:
```
[15:49:37] - WARNING: Socket not open for unit 1. Attempting to reconnect...
[15:49:37] - WARNING: Socket not open for unit 2. Attempting to reconnect...
[15:49:37] - WARNING: Socket not open for unit 3. Attempting to reconnect...
```

incomplete modbus messages:
```
[15:49:37] - ERROR: Error reading temperature from unit 2: Modbus Error: [Input/Output] Modbus Error: [Invalid Message] Incomplete message received, expected at least 4 bytes (3 received)

frequently no response from slave controller:[15:49:38] - ERROR: Error reading temperature from unit 2: Modbus Error: [Input/Output] No Response received from the remote slave/Unable to decode response

### DIFF
--- a/instrumentctl/E5CN_modbus.py
+++ b/instrumentctl/E5CN_modbus.py
@@ -88,8 +88,8 @@ class E5CNModbus:
         """
         try:
             if self.client.is_socket_open():
-                self.log("Modbus client already connected.", LogLevel.DEBUG)
-                return True
+                self.client.close()
+                time.sleep(0.1) # for socket cleanup
 
             if self.client.connect():
                 self.log(f"E5CN Connected to port {self.client.comm_params.port}.", LogLevel.INFO)
--- a/instrumentctl/E5CN_modbus.py
+++ b/instrumentctl/E5CN_modbus.py
@@ -88,8 +88,8 @@ class E5CNModbus:
         """
         try:
             if self.client.is_socket_open():
-                self.client.close()
-                time.sleep(0.1) # for socket cleanup
+                self.log("Modbus client already connected.", LogLevel.DEBUG)
+                return True
 
             if self.client.connect():
                 self.log(f"E5CN Connected to port {self.client.comm_params.port}.", LogLevel.INFO)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mslaffin** [comment]
I incorporated several changes and tested this on 11/25. These changes included:
- Added `is_initialized` event flag 
- thread locking around connection methods
- added retry mechanism (set `retries=2` in client initialization)
- added connection validation before starting temperature reading threads
- added small delays after connection operation

many of the modbus connection errors previously seen are now resolved, however the logs are still intermittently showing incomplete messages like:
```
Modbus Error: [Invalid Message] Incomplete message received, expected at least 4 bytes (3 received)
```

These incomplete message errors occur across all three units but recover automatically. This does not appear to affect the core functionality of the temperature reading/display. The 500ms polling cycle might be a bit aggressive, although the E5CN datasheet claims it can achieve cycles of down to 60ms..

I've attached a logfile that shows this: [log_2024-11-25_17-38-27.txt](https://github.com/user-attachments/files/17926402/log_2024-11-25_17-38-27.txt)

Another problem has popped up though, and I think it requires more investigation before this branch can be merged into develop. The following serial port access conflict messages appear in the terminal (not in the Message & Errors frame) at approximately the same rate as the "incomplete message" errors:
```
could not open port 'COM4': PermissionError(13, 'Access is denied.', None, 5)
```

This is likely related, but it's unclear to me exactly how at this point, and why it's not being redirected to the logs.

bwalkerMIR:
I looked at your log file and it looks like you're still getting plenty of good messages and like you said it's an occasional error that occurs maybe ever 10 reads on average or so. If the code handles this error gracefully, I think we should make a task in SOFT Construction to follow up on it later but should not prevent closeout for CCS QAI at this point since it functions.

I agree the error of denied access to the COM port is more problematic and should be checked out before merging.

mslaffin:
BW and I decided that we'll resolve the PermissionError bug in another PR since it doesn't affect the core functionality, so I'm going to merge this now.

PR #26, opened 2024-11-17
Feature/comportdetection

mark11778:
Have tested to make sure this works in all expected cases with the G9. 

On a side note, the error/alarm red flashing light turns off once the connection is reestablished, sorry Brandon.

Will need to add the functionality for the VTRX subsystem and the Cathode Heating.

This does include a lot of the G9 branch files, let me know if I should remove them from this pull request


mark11778:
Oh, if you keep in mind that when this branch is merged into develop you will have to remove usr/com_ports.json so the handle the new format.

mslaffin:
This looks pretty solid.

I think we first have to clean this branch so that we're not merging outdated G9Driversim changes into develop. Makes it hard to review the detection/config storages changes in isolation.
`git diff develop...feature/comportdetection`

the easiest way might be:
- create a clean branch starting from develop
- cherry-pick only the comportdetection changes
- basically delete this one

but if there's too much dependency, then we could wait for the G9 branch to be merged to develop, then rebase this on the updated develop.

Also need an attachment of the log file

instrumentctl/g9_driver.py
```diff
@@ -86,12 +82,20 @@ def _setup_serial(self, port, baudrate, timeout):
                     bytesize=serial.EIGHTBITS,  
                     timeout=timeout  
                     ) 
+                
+                self._lock = threading.Lock()
+                self._response_queue = queue.Queue(maxsize=1)
+                self._running = True
+                self._thread = threading.Thread(target=self._communication_thread, daemon=True)
+                self._thread.start()
```
mslaffin, line 90: @mark11778 these changes affect the g9_driver.py but aren't reflected in the G9 driver PR. 

Can you move these to [PR#16](https://github.com/uw-loci/EBEAM_dashboard/pull/16) and paste in a fresh log file to the comments (without committing the logfile) so that I can merge it?

mark11778: Ope, sorry I missed this I will make a branch now

mark11778:
no longer works, so switching this to draft to figure out what broke

mark11778:
no longer relativent, and most code made it into develop in [PR#54](https://github.com/uw-loci/EBEAM_dashboard/pull/54)

PR #28, opened 2024-11-21
Tkinter thermometer graphic

mslaffin:
This PR replaces the environmental subsystem matplotlib figure with a lightweight Tkinter canvas widget:
Screenshot 2024-11-20 204548
Screenshot 2024-11-20 204433


bwalkerMIR:
Nice! This would be much faster right?

mslaffin:
> Nice! This would be much faster right?

yea absolutely! matplotlib creates a full plotting environment that has a lot of overhead which should all be cut out now

PR #34, opened 2024-12-03
PMON Driver

mslaffin:
Opening this up for discussion of initial work of the ProcessMonitor subsystem and DP16 driver.

Hardware: [DP16PT-330-C24](https://www.newark.com/omega/dp16pt-330-c24/digital-panel-meter-1-16-din-240v/dp/31AC5835)
Communication Manual: [PMON digital meter RS-485 communication manual.pdf](https://github.com/user-attachments/files/17987276/PMON.digital.meter.RS-485.communication.manual.pdf)


mslaffin:
image

mark11778:
image

mark11778:
image

appendix C page 53  RS-485 communication manual

mark11778:
image

**Big-endian**

mslaffin:
Just so everyone is aware, we shouldn't have to worry about explicitly dealing with byte structure or CRC calculations when using [pymodbus](https://pymodbus.readthedocs.io/en/latest/).

These low level requirements are already implemented at various levels by pymodbus. For example, the FramerRTU class handles the CRC check [here](https://github.com/pymodbus-dev/pymodbus/blob/cadbec722c5f57d1fff22a57e2bfff3697fd2341/pymodbus/framer/rtu.py#L119)

mark11778:
Ope thanks for the heads up, Ill have to look into this library

mslaffin:
I'm thinking we'll want a method to interpret the scaling of the "process value" that we're reading from register 39. They provide register 8 (RDGCNF) which I believe stands for "Reading Configuration" and can be used to determine this decimal place. 
image

They have this format table in section 5.7.2 (iSeries protocol settings), but don't have a corresponding table in the Modbus section. I think this is consistent across protocols.

This shows the decimal position is encoded in the last 3 bits of RDGCNF:

image

mslaffin:
just noticed the mapping of these process monitors appears to have changed. Originally I had:
```python
      self.thermometer_map = {
          'Solenoid 1': 1,
          'Solenoid 2': 2,
          'Chamber Bot': 3,
          'Chamber Top': 4,
          'Air temp': 5
      }
```
based on this labview concept:
Capture_labview

but per the 2024-06-21 hardware prod spec, the order of Chamber Top and Chamber Bottom have been swapped. I'm going to update the mapping to match this spec:
image

bwalkerMIR:
Oh nice catch on the mapping. The PMON Hardware Product Spec is correct.

mslaffin:
guys I think I understand why this is still slowing down the GUI so much.

Each unit is getting its own dedicated thread that loops indefinitely and reads data every 0.2 seconds. But I don't think this is actually parallelizing the work as we intend and could be slowing things down even unnecessarily.

The Modbus communication is gated by a single serial connection and a single `modbus_lock`. With multiple threads, we could be effectively serializing access to the Modbus client, which negates the benefit of having multiple simultaneous threads.

I think we could revise our threading scheme to use only one thread. 

We could have the worker thread periodically poll _all units_ in sequence, opposed to starting one thread _per unit_. I think this could help us avoid those threads contending for the lock every 200 ms.

mslaffin:
At this point, I think this branch is safe to merge into develop and shouldn't pose a risk to the existing subsystems. I'm going to merge and close this but I'll open up a follow-up PR for clean-up and any changes that might come up during testing.

PR #40, opened 2024-12-30
CCS-misc-repairs

mslaffin:
This PR is intended for CCS rework related to issues observed during testing on 12/26:
- Increase minimum y-axis resolution for CCS temp plots
    - currently defaults to a range just outside measurement, so small variations cause large distortions
-  Intermittent 0.0V display glitch from 9104 power supplies (observed on all three, likely parsing problem)
    - messes up recorded PS data. Glitch is observed on GUI, not on physical display panel
- Overvoltage limit setting failure when using `Config` tab user entry field
    - This configuration succeeds during `initialize_power_supplies`. need to add formatting checks on user input prior to sending


bwalkerMIR:
@mslaffin fyi @theoreotically thinks he found the bug with failuing to set OVP and OCP and he will desicribe it more here and push a fix

bwalkerMIR:
Also @theoreotically found this:

BUG: It is possible to cause an error that can only be fixed by manual intervention if the following is done:

Set heater voltage/current such that value <= OVP or value <= OCP
Change OVP or OCP to a value less than voltage or current.
Switch output on. (i.e, right now, voltage is set to a value higher than OVP, which should not be allowed. 

The power supply throws and error. We should have a check that does not allow us to switch on the output

theoreotically:
> @mslaffin fyi @theoreotically thinks he found the bug with failing to set OVP and OCP and he will desicribe it more here and push a fix

There was a mismatch between how we dealt with overvoltage (OVP) & overcurrent (OCP) during initialization. I have changed the relevant code and functions to reflect the same.

Now initialization and editing of OVP & OCP are same. However, OCP currently has no bounds checks at all, meaning that OCP can be set to any value. The same appears to be true for voltage and current, while OVP is the only (from what I have come across) that has a bounds check. Will elaborate further on Asana.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -98,6 +106,16 @@ def set_voltage(self, preset, voltage):
...
+        
+    def validate_voltage(self, voltage):
+        """Check if the voltage is less than the OVP."""
+        ovp = self.get_over_voltage_protection()
+        if ovp is None:
+            self.log("Could not validate voltage. OVP unavailable.", LogLevel.ERROR)
+            return False
+        if voltage >= ovp:
+            self.log(f"Voltage {voltage:.2f}V is greater than OVP {ovp:.2f}V", LogLevel.ERROR)
+            return False
```
mslaffin, line 118: It looks like there's a bug in this `validate_voltage`. There's no path where it returns True?

This seems like it would always return False to the caller...

README.md
```diff
@@ -171,7 +171,7 @@ Create a virtual environment:
 ```
 python -m venv venv
 ```
-Activate the virtual environment (assuming on Windows):
```
bwalkerMIR, line 174: example review by BW

bwalkerMIR: I will resolve this.

bwalkerMIR (approved):
ML and I think these changes look great.

PR #44, opened 2025-01-05
Improvement/vtrx proportional resize

mslaffin:
Front-end changes to proportionally resize indicators and text positioning

### DIFF
--- a/subsystem/vtrx/vtrx.py
+++ b/subsystem/vtrx/vtrx.py
@@ -132,10 +132,21 @@ class VTRXSubsystem:
                 time.sleep(1)
 
     def _create_indicator_circle(self, parent_frame, color="grey"):
-        """ Switch state circular indicator on a canvas. """
         canvas = tk.Canvas(parent_frame, width=30, height=30, highlightthickness=0)
+        canvas.bind('<Configure>', lambda e: self._resize_indicator(canvas, e))
         oval_id = canvas.create_oval(5, 5, 25, 25, fill=color, outline="black")
         return canvas, oval_id
+
+    def _resize_indicator(self, canvas, event):
+        canvas.delete('all')
+        width, height = event.width, event.height
+        margin = min(width, height) // 4
+        canvas.create_oval(
+            margin, margin, 
+            width - margin, height - margin, 
+            fill=canvas.itemcget(canvas.find_all()[0], 'fill') if canvas.find_all() else 'grey', 
+            outline="black"
+        )
 
     def process_queue(self):
         try:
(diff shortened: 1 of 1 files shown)


mslaffin:
I've expanded this PR to add time window configurability for the pressure plot, as well as an export mechanism to save off the plot figure as a .png.

Currently, the selection dropdown defaults to 5 minutes, but this window can be increased all the way to 10 hours:
Animation
 
Prior to this, BW and I had discussed an issue where pressure variations at high vacuum were not observable in the plot due to the statically defined y-axis limits. This sort of defeated the utility of the plot itself, but I believe this can be resolved in this PR.

If we started a pump-down procedure from 1 atm with a 5-hour time window, this PR code would exhibit the same problem. However, if we then switch to a 5-minute window, the `update_time_window` method will filter out the older data points and keep only the previous 5 minutes. Then `update_plot` would recalculate the y-axis limits based off only the recent data. 
- for a vacuum around `10^-6`, this would zoom in to roughly `5*10^-7` to `2*10^-6`
- should be able to see small pressure variations with these new axis limits during high vac state


The problem that still exists here is that when changing to a short time window (i.e. 5 minutes), the code filters out all data points older than 5 minutes. If we were then to switch back to 5 hours from 5 minutes, we would then have a mostly empty plot area except for the last minutes prior to switching. 

I think it would be possible to maintain a full history separate from what is displayed in the plot. This would be bounded by the max available time window (10 hours, or w/e we decide upon), and instead of permanently filtering data when changing windows, we could just update the portion of the full history that is displayed.

Before I do this though, I want some confidence in how much memory overhead this buffering would incur (10 hours of samples every 500ms, each sample containing 1 `float` and one python `datetime` object)

mslaffin:
I've finished the pressure history configuration feature and have performed some testing with a mock VTRX interface. 

This now stores up to 10 hours of pressure readings and allows you to adjust the time window in and out. Entries older than MAX_HISTORY_SECONDS are pruned.

I estimate that this can increase the app's memory overhead by 5 MB or so over a full 10 hour session. 
Each element in the history requires a python datetime object and pressure float. Here are the sizes:
```
>>> import sys
>>> import datetime
>>>
>>> sys.getsizeof(datetime.datetime.now())
48
>>>
>>> sys.getsizeof(float())
24
```
So storing 72 bytes every 500ms for 10 hours comes out to approximately 5MB (5184000bytes)
restored_history

Y-axis limits automatically adjust when viewing a steady reading:
autoscaling_y-axis

bwalkerMIR:
This is awesome!

PR #51, opened 2025-01-24
CI pipeline testing pipeline config along with tests for G9

mark11778:
Updated requirements.txt,
workflow dir with the yml file
added ci tests for G9

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -1,0 +1,28 @@
+name: Python Tests
+
+on:
+  push:
+    branches: [main, develop, feature/citesting]  
+  pull_request:
+    branches: [develop, main]
+
+jobs:
+  test:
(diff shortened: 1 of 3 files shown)


.github/workflows/python_tests.yml
```diff
@@ -0,0 +1,28 @@
+name: Python Tests
+
+on:
+  push:
+    branches: [main, develop, feature/ciG9]
```
mslaffin, line 5: @mark11778 I think this PR looks pretty much good to go.

It seems unnecessary to have `feature/ciG9` included in the push triggers though. 

This is a temporary branch and it seems like we shouldn't want that hard-coded here because it'll be deleted after merging?

Would it cause problems to remove this?

mark11778: Ope, good catch. Yeah that was just for testing. Ill remove it now

PR #53, opened 2025-02-03
Feature/graceful exit

AryanSV5:
This is more of a graceful-start feature rather than graceful-exit. Earlier we had trouble re-running application as some subsystems like PMON and SIC weren't displaying any readings. This usually occurred when we tried to run the application with all the connected subsystems and had to re-start the laptop itself to re-establish connection with the subsystems. 

I just made a cleanup method in dashboard.py which calls the closed_com_ports method for every subsystem and closes all open com ports. This is reported in the log file which is accessible after quitting the application to confirm which specific ports have been closed.

### DIFF
--- a/instrumentctl/G9SP_interlock/g9_driver.py
+++ b/instrumentctl/G9SP_interlock/g9_driver.py
@@ -120,7 +120,7 @@ class G9Driver:
             except Exception as e:
                 self.log(f"Communication thread error: {str(e)}", LogLevel.ERROR)
                 #TODO: this might be solved with the comport detection, but if not might what to define this somewhere else
-                self._response_queue.queue[0] = ([0] * 13, [0] * 13, 0)
+                #self._response_queue.queue[0] = ([0] * 13, [0] * 13, 0)
                 time.sleep(0.5) # back off on errors
                 
             time.sleep(0.1)  # minimum sleep between successful reads
--- a/dashboard.py
+++ b/dashboard.py
@@ -55,6 +55,7 @@ class EBEAMSystemDashboard:
     def __init__(self, root, com_ports):
         self.root = root
         self.com_ports = com_ports
+        self.stop_threads = False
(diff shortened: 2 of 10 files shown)


instrumentctl/G9SP_interlock/g9_driver.py
```diff
@@ -0,0 +1,378 @@
...
+                            self._response_queue.get_nowait()
+                        except queue.Empty:
+                            pass
+                                       
+                        self._response_queue.put(result)
+
+            except Exception as e:
+                self.log(f"Communication thread error: {str(e)}", LogLevel.ERROR)
+                #TODO: this might be solved with the comport detection, but if not might what to define this somewhere else
+                #self._response_queue.queue[0] = ([0] * 13, [0] * 13, 0)
```
mark11778, line 123: line 123 - which was determined to be throwing the error, was added for when an exception was thrown all the indicators would be shown as red as opposed to their last message reading.

it was an index out of bounds error

mslaffin:
@AryanSV5 can you change the base branch for this PR from `main` to `develop`?

instrumentctl/G9SP_interlock/g9_driver.py
mslaffin: I am thinking that this Index out of bounds might've been due to direct accessiing `queue.queue[0]` when the queue is empty.

Looks like we should actually be using [put()](https://docs.python.org/3/library/queue.html#queue.Queue.put) instead
Something like `self._response_queue.put(([0] * 13, [0] * 13, 0))`

mslaffin: Actually, I think in this case it would make sense to also clear any existing items in the queue prior to doing `_response_queue.put(([0] * 13, [0] * 13, 0))`

mslaffin (approved):

PR #54, opened 2025-02-07
G9SP Response Queue IndexError Fix

mslaffin:
Context from [PR#53](https://github.com/uw-loci/EBEAM_dashboard/pull/53):
image

I believe this was effectively the error that was happening:
```
>>> import queue
>>>
>>> q = queue.Queue()
>>> item = q.queue[0]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: deque index out of range
>>>
```


bwalkerMIR:
This error has showed up a lot!!

AryanSV5:
Tested branch in lab on 2/10/2025-11:39am and the index error no longer comes up. Neither the console or logs displays the error. Ran tests with SIC individually and all also with other subsystems.

mark11778:
It showed up again, but after ^ those commits, no longer being output to log ... but on disconnection the program stops responding

mslaffin:
hey @mark11778 looks like you caught another instance of a direct-access of `.queue[0]`. I have some ideas about what's happening. 

Your first change `e37df8c` changed to `get_nowait()`
```
- return self._response_queue.queue[0]
+ return self._response_queue.get_nowait()
```
this seems problematic because get_nowait actually [removes the item from the queue](https://docs.python.org/3/library/queue.html#queue.Queue.get), so each time we check the status, this actually consumes/removes the data instad of just reading the latest status.

Then it looks like you reverted back to looking at .queue[0] but with an empty check first. I think this sort of reintroduced the original thread-safety problem. The fact that the program hangs on disconnect is probably due to the GUI thread blocking because we're still trying to access the queue unsafely..

I'll push some changes to `get_interlock_status` that aligns this with the pattern Aryan tested in `_communication_thread`

mark11778:
Ope, I missed your comment from before, but I was actually talking to Brandon about the same thing. I too was saying that I think it is an issue with how we are handling the threading locks.

This might also problem that the DP16 also shares to some extent

mark11778:
After some of the changes on how the bytes are being read in, the indicators now, show disconnection basically instantly, for reconnection(with only the SIC running and connected) it takes 6-7s. 

But, no longer throwing index errors or crashing the GUI

instrumentctl/G9SP_interlock/g9_driver.py
```diff
@@ -169,54 +188,70 @@ def _read_response(self):
...
-            data = self.ser.read_until(self.FOOTER)
-
-            if len(data) != self.EXPECTED_DATA_LENGTH:
-                length_error_msg = f"Invalid response length: got: {len(data)}, expected 199 bytes"
-                self.log(length_error_msg, LogLevel.ERROR)
-                raise ValueError(length_error_msg)
+            s, t = time.time(), .5
+            data = bytearray()
+            while time.time() - s < t:
+                chunk = self.ser.read(50)
```
mslaffin, line 194: Hey @mark11778, nice work with this chunked reading approach, I think this is a really good idea. 

It's not clear to me how you settled on 50 bytes at a time though? I don't see this matching any of the specified data segments in the G9 manual. 

image


Seems to me like this requires 199bytes/50 = 4 chunks to get a full response message. What are your thoughts on sizing the chunks based on specified field lengths? Was there some buffer overflow concern with the higher data byte length?

I'm thinking something like:
```python
header = self.ser.read(7)
if not header or header[0:1] != self.ALWAYS_START_BYTE:
    return None

# Read data section (188 bytes)
data = self.ser.read(188)
if not data or len(data) != 188:
    return None

# Read checksum and footer (4 bytes)
footer = self.ser.read(4)
if not footer or footer[-2:] != self.FOOTER:
    return None

return header + data + footer
```

mark11778: Yeah, so with the 9600 which means we have can receive 9600/10 = 960 bytes/s, and if we are pull at every 0.05s we can expect to get 48 bytes at each time.

ans ser.read(50) will read up to 50 bytes(and return whatever is finds), so like the bytes are available it will be read in 4 cycles, otherwise it will take 5 cycles.

I see some issue with suggested code
```
header = self.ser.read(7) # if this returns a error package
if not header or header[0:1] != self.ALWAYS_START_BYTE:
    return None

# Read data section (188 bytes)
data = self.ser.read(188) # given if a error package was sent this would contain all  
if not data or len(data) != 188:  # but it would make this true
    return None # then this would return None

# Read checksum and footer (4 bytes)
footer = self.ser.read(4)
if not footer or footer[-2:] != self.FOOTER:
    return None

return header + data + footer
```
The problem that I was seeing with the larger reads was that on disconnection, I think the serial connection would just wait for a response, that it was never going to get, so it would lead to the GUI to stop responding?

mslaffin: Oh I see now. That makes a lot of sense!
Let's go with your implementation.

mark11778:
Should be ready to merge into develop now

mark11778 (approved):

PR #55, opened 2025-02-17
CCS bugfix errata: Fixing small but significant issues

theoreotically:
An aside to fix tiny bugs within CCS. This may also be used to fix issues identified during 14th Feb testing

### DIFF
--- a/subsystem/cathode_heating/cathode_heating.py
+++ b/subsystem/cathode_heating/cathode_heating.py
@@ -41,9 +41,9 @@ class CathodeHeatingSubsystem:
     MAX_POINTS = 60  # Maximum number of points to display on the plot
     OVERTEMP_THRESHOLD = 200.0 # Overtemperature threshold in C
     ERROR_COLORS = {
-        'normal': 'blue',          # Normal operation
+        'normal': 'blue',         # Normal operation
         'overtemp': 'red',        # Overtemperature condition
-        'ERROR': '#FFA500',  # Communication error
+        'ERROR': '#FFA500',       # Communication error
         'DISCONNECTED': '#808080'
     }
     
@@ -89,11 +89,11 @@ class CathodeHeatingSubsystem:
         self._init_config_variables()        # Configuration and safety settings
 
         # System initialization sequence
-        self.init_cathode_model()            # Initialize cathode physics models
-        self.setup_gui()                     # Set up graphical interface
-        self.initialize_temperature_controllers()  # Connect to temperature controllers
-        self.initialize_power_supplies()      # Connect to power supplies
-        self.update_data()                   # Start the data update loop
+        self.init_cathode_model()                   # Initialize cathode physics models
+        self.setup_gui()                            # Set up graphical interface
+        self.initialize_temperature_controllers()   # Connect to temperature controllers
+        self.initialize_power_supplies()            # Connect to power supplies
+        self.update_data()                          # Start the data update loop
 
     def _init_prediction_variables(self):
         """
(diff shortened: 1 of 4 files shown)


mark11778:
Sorry miscommunication still needs to be tested

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -592,36 +623,36 @@ def initialize_power_supplies(self):
                         self.log(f"Asserted preset mode 3 for cathode {cathode}. Response: {get_preset_response}", LogLevel.INFO)
 
                     # Set and confirm OVP
-                    ovp_value = int(round(self.overvoltage_limit_vars[idx].get() * 100)) # TODO: Do we need to round?
-                    self.log(f"Setting OVP for cathode {cathode} to: {ovp_value / 100:.2f}", LogLevel.DEBUG)
+                    ovp_value = int(round(self.overvoltage_limit_vars[idx].get()))
+                    self.log(f"Setting OVP for cathode {cathode} to: {ovp_value:.2f}", LogLevel.DEBUG)
                     if ps.set_over_voltage_protection(f"{ovp_value:04d}"):
```
mslaffin, line 628: Looks like there's inconsistency in the parameter types for setting OVP values. 

Here in `initialize_power_supplies` the ovp value is passed as an integer formatted as a 4-digit string, but in `set_overvoltage_limit` it's formatted as a float with 4 decimal places (line 742):
```python
ovp_set_response = self.power_supplies[index].set_over_voltage_protection(f"{raw_value:.4f}")
```

The `set_over_voltage_protection` in the PowerSupply9104 class should already handle conversion of any input (string or number) to a float, multiply by 100, round to an integer, and format as a 4-digit string:
```python
def set_over_voltage_protection(self, ovp_volts):
    """Set the over voltage protection value."""
    """ Expected response: OK[CR] """
    ovp_centivolts = int(round(float(ovp_volts) * 100))
    command = f"SOVP{ovp_centivolts:04d}" # format as 4-digit string
    response = self.send_command(command)

    if response and "OK" in response:
        return True
    else:
        self.log(f"Failed to set OVP to {ovp_centivolts:04d}", LogLevel.DEBUG)
        return False
```
I'm having a hard time understanding how this double conversion wouldn't be problematic for setting values from at least one of these spots. Seems like it would be simpler to just let the PowerSupply9104 class handle all the formatting and change CathodeHeatingSubsystem to pass in the raw values directly?

theoreotically: > Seems like it would be simpler to just let the PowerSupply9104 class handle all the formatting and change CathodeHeatingSubsystem to pass in the raw values directly?


That was the intention. I believe I missed updating a few values from within CathodeHeatingSubsystem while debugging the threading lock issue simultaneously. Prior to my updates, CathodeHeatingSubsystem was preformatting all values and passing it over to PowerSupply9104. Since this wasn't consistent, with how other functions passed values, the intention was to have 9104 deal with strings and numbers being passed in. Will make the appropriate changes and test during hours on Monday.

As mentioned on Asana,

> I would like to test a few more changes of PR 55 before I mark it for a merge review as the code streamlining and tool-tip inclusions had initially broken some stuff (they work now).

the intention here was to go over and test CathodeHeatingSubsystem again before marking the PR for review.

theoreotically:
@mslaffin @bwalkerMIR I have fixed the issues within OVP and OCP identified earlier and have found part of the reason for Voltage Ramp's weird* behavior. CCS is stable during ramps now and does not cause any crashes. I would like to merge these changes into develop. Will open a new PR post-audit to fix the weird Voltage ramp* behavior.

*weird voltage ramp behavior refers to the behavior observed in the [videos from testing](https://drive.google.com/drive/folders/1BYUVeKz0re7bVIToUGwMVsjhI2zzWFBP?usp=drive_link) on 2025-03-05 where voltage on PS display is initially at voltage set, then decreases until a value and increases to the voltage set. Turns out we are setting voltage to voltage set first, followed by a ramp from 0 to voltage set. We should only be doing the ramp. This and other bugs should be fixed once code audit is done and refactor recommendations are discussed.

mslaffin (approved):
hey @theoreotically I've looked through these changes and everything looks good and consistent to me! You've got `PowerSupply9104` handling the formatting internally, and `CathodeHeatingSubsystem` passing in the raw data (after validation)

PR #56, opened 2025-02-27
Feature/machine progress bar

AryanSV5:
Added a machine status progress bar to the dashboard by creating another row. I made this subpanel in utils.py and check for updates from the interlocks and pmon subsystems to determine if they are active or not. 

image


dashboard.py
```diff
@@ -278,24 +285,33 @@ def create_subsystems(self):
...
                 com_ports = self.com_ports['Interlocks'],
                 logger=self.logger,
-                frames = self.frames
+                frames = self.frames,
+                active = self.machine_status_frame.MACHINE_STATUS
             ),
             'Oil System': subsystem.OilSubsystem(
                 self.frames['Oil System'],
-                logger=self.logger
+                logger=self.logger,
```
mslaffin, line 293: I wonder if we should add a `MACHINE_STATUS` for the oil system? And for the VTRX system? 

It seems like that would give us consistency across all subsystems at basically no cost. As long as we can fit the text.

utils.py
```diff
@@ -345,3 +345,74 @@ def hide_tip(self):
...
+
+    def update_status(self, status_dict=None):
+        """
+        Update the status of each machine indicator.
+        :param status_dict: Dictionary with status names and their boolean state.
+        """
+        if status_dict is None:
+            status_dict = self.MACHINE_STATUS
+        self.update_labels(status_dict)
+        self.parent.after(self.update_interval, self.update_status)  # Schedule the next update
```
mslaffin, line 412: this update method is getting called repeatedly even when there are no changes. We'll probably want to track if any status in the dict has actually changed before scheduling the update.

mslaffin (commented):
Nice work, I think this looks clean

dashboard.py
bwalkerMIR: Yes Arundhati and I talked about this just now and think this is a great idea

utils.py
bwalkerMIR: Awesome it looks like you fixed this in commit 4d3af84

PR #57, opened 2025-03-01
Feature/taller machine status and allow fixed height as only set dimension

bwalkerMIR:
The pull request slightly increases the height of the Machine Status bar at the bottom of the GUI, which was too small before and the text was running off a bit.

Additionally, it looks like the previous develop branch code did not allow you to set a fixed height for a frame while allowing no dimension on the width (so it would still stretch across the screen). This PR fixes the logic to allow setting height alone or width alone without having the other set.

image

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -34,7 +34,7 @@ frames_config = [
     ("Cathode Heating", 4, 980, 450),
 
     # Row 5
-    ("Machine Status", 5, None, 50)
+    ("Machine Status", 5, None, 55)
 ]
 
 class EBEAMSystemDashboard:
--- a/dashboard.py
+++ b/dashboard.py
@@ -34,7 +34,7 @@ frames_config = [
     ("Cathode Heating", 4, 980, 450),
 
     # Row 5
-    ("Machine Status", 5, None, 55)
+    ("Machine Status", 5, None, 42)
 ]
 
 class EBEAMSystemDashboard:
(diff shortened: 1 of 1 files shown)


mslaffin (commented):
I think there's a deeper problem that led to this appearance. When I opened this branch on windows 10 on my 3840x2160 display at 175% scaling, the status dividers do not fully span the row height:
Capture4


We should be using a proportion-based layout instead of these fixed dimensions (`50`->`42`) , because it appears differently depending on the users' screen resolution. I've got this resolved in a new commit. 
175% resolution w/fix:
Capture2_fixed

225% resolution w/fix:
Capture3_fixed
next steps will involve adjusting the text size ideally

bwalkerMIR:
Interesting. Original shown first:
image
Michael's new commit does make it better with relative scaling, but the whole thing doesn't fit on the screen on my computer. It seems to be an issue when running on high resolution screens. New commit from Michael: 
image

bwalkerMIR:
If I change the DPI settings of python on my laptop, the CCS area and text resolution is much better overall  but PMON is way off as well as button sizes etc:
image

bwalkerMIR:
Closed this PR because there is a bigger issue as Michael mentioned with high DPI screens, so I don't want to push a partial solution on an older branch when the current develop works well on the main experiment laptop. It just doesn't work well for development on high DPI laptops. I'll make a new task in asana to figure out what the problem is with running on high DPI screens.

PR #58, opened 2025-03-12
Proposed CCS flow

theoreotically:
Updates CCS flow to correctly only set voltage to power supply due to Controlled Voltage mode. The new flow charts incorporate the toggle button and proposed Voltage LUT. Note: The proposed Voltage LUT should be accounted for in updated functions where user sets target current. For a user setting a voltage, we can be implicit about this or explicit (and update the flowchart accordingly)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210465931584534


theoreotically:
Note: Based on the flowcharts, a sleep(0.1) was updated to sleep(0.5) as the 100ms sleep seemed arbitrary compared to the known 500ms communication timeout. Idle State Monitoring flowchart, if correct, could contain issues as split threads may trying writing to a disabled/destroyed object. It is recommended that in a different PR, threads across the platform are reviewed by project lead to ensure no actual race conditions (May identify and fix some lasting GUI quitting issues too)

theoreotically:
Will have to fix 0.58V bug in future, but for now, there's already a bunch of changes that are slightly overreaching the PR's initial goal.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1295,36 +1290,45 @@ def set_target_current(self, index, entry_field):
...
+                    
+                    if current_ovp is None:
+                        self.log(f"Unable to get current OVP for Cathode {['A', 'B', 'C'][index]}. Aborting voltage set.", LogLevel.ERROR)
+                        return
+                    if heater_voltage > current_ovp:
+                        self.log(f"Calculated voltage ({heater_voltage:.2f}V) exceeds OVP ({current_ovp:.2f}V) for Cathode {['A', 'B', 'C'][index]}. Aborting.", LogLevel.WARNING)
+                        msgbox.showwarning("Voltage Exceeds OVP", f"The calculated voltage ({heater_voltage:.2f}V) exceeds the current OVP setting ({current_ovp:.2f}V). Please adjust the OVP or choose a lower target current.")
+                        return
+                    # Set the voltage and current on the power supply
+                    voltage_set_success = self.power_supplies[index].set_voltage(3, Decimal(heater_voltage))
```
mslaffin, line 1318: I guess I'm not convinced why we would want to remove the explicit current limit setting here..

Originally we hada dual constraint situation where the ps is commanded to output a specific voltage, and is also given a max current limit for that voltage point. The success checks verify both parameters were accepted by the hardware.

Wouldn't this change mean we're now rely on the global OCP limit rather than a specific Upper Current Limit? This seems risky because the power supply could deliver much more current than is appropriate for a particular ES440 voltage operating point.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1382,13 +1386,13 @@ def reset_power_supply(self, index):
...
         Side effects:
-            - Sets voltage and current to 0
+            - Sets voltage to 0 and current to 9.0
             - Resets all prediction variables to '--'
             - Logs the reset action
         """
         if self.power_supply_status[index]:
             self.power_supplies[index].set_voltage(3, 0.0)
-            self.power_supplies[index].set_current(3, 0.0)
+            self.power_supplies[index].set_current(3, 9.0)
```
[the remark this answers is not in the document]
theoreotically: After having run tests and discussed how the power supply works with Brandon, it appears that given a fixed resistance, if the power supply targets a given voltage, if the allowable current is close to the current resulting from the given resistance and voltage, a current limit scenario arises. The plan was to not repeatedly set voltage and current (V = IR, setting both V and I didn't seem right). The 9.0 is based on the LUT as 1.0V has a current over the initial limit of 8.5A.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1251,11 +1251,6 @@ def set_target_current(self, index, entry_field):
             - Updates heater voltage display
             - Logs actions and any errors
         """
-        if self.toggle_states[index]:
-            # if the output toggle is enabled, show a warning message
-            msgbox.showwarning("Warning", "Disable the output before setting a new target current.")
-            return
-
```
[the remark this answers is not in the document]
theoreotically: Brandon asked for the ability for users to be able to reset or change a target without having to disable the output. It was deemed appropriate given that in the event that a user enters an incorrect target, its less stressful on the cathode if the user simply changes the target and we ramp a small difference compared to toggling supplies, unnecessarily cycling the cathodes through two major ramps.

subsystem/cathode_heating/cathode_heating.py
theoreotically: I understand the intention, but the power supply limits itself given a current setting due to some inaccurate internal meters in the supply causing issues. So, if this were something that we wanted to implement, it would be necessary to give the current setting some headroom over what we expect. In its current implementation, without the documentation, it wasn't clear why both a voltage and current were being set every time. This check can be reverts given that we decide  on a reasonable "current headroom".

subsystem/cathode_heating/cathode_heating.py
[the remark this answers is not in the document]
mslaffin: I understand this adds convenience for adjusting the setpoint during operation. 

It still seems to go against the standard practice of separating the "setting" phase from the "energizing" phase for controlling sensitive loads, which is why I had the safety check. 

Instead of bypassing this fully, we could open a dismissible message box to warn the action carries risk? 

That would be different from the original hard requirement to fully disable through toggle off. 

> its less stressful on the cathode if the user simply changes the target and we ramp a small difference compared to toggling supplies

I'm still not fully convinced here. There's no ramp control for mid-operation `set_target_current` changes like there is in `toggle_output`, it just [sets the voltage](https://github.com/uw-loci/EBEAM_dashboard/blob/53b431a300b9117c9f4b971f4022a35993bb38e9/subsystem/cathode_heating/cathode_heating.py#L1318). There are also only limited input bounds checks, so this means the user bears responsibility to keep the difference within a safe adjustment window (varying across vacuum, temp conditions)?

subsystem/cathode_heating/cathode_heating.py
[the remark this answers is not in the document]
mslaffin: > given a fixed resistance

The ES-440 V-I curves show that the resistance is not fixed. The resistance changes significantly with current and temperature. 

> The plan was to not repeatedly set voltage and current (V = IR, setting both V and I didn't seem right).

Setting both V and I doesn't seem redundant or contradictory to me. The 9104 is can accept both parameters simultaneously. One parameter (V) establishes the desired operating point and the other (I) provides a safety boundary to prevent delivering power outside of the characterized range. 

If specific operating points require higher currents, then the proper solution should be as you've said to add a calculated headroom factor when setting operating points, not to permanently raise the current limit in the reset state.

And besides this, the reset function should always place the system in the most de-energized state (0.0A), not its maximum allowable state (9.0A), right?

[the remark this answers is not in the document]
theoreotically: I believe the issue might have risen from us focusing on the fixed 0.2 Ohm test resistor as opposed to the ES-440. Will revert changes related to the IV settings and reset state. Setting only the voltage would be more of a nice-to-have for when we were working on the 0.2 Ohm resistor. 

I imagine adding the headroom may result in other issues, so the only real way to know if we need to add the headroom is to test with a resistor with similar IV characteristics to our cathode (or at least one with a non-linear curve).

subsystem/cathode_heating/cathode_heating.py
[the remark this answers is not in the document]
bwalkerMIR: I think we should be able to change the voltage during operation without having to turn off the output first. If we change the voltage while the output is on, why don't we just have it also ramp up or down if the ramp is toggled on? If the ramp is toggled off, then it would be immediate and up to the user, but I still don't think we really need a warning.

subsystem/cathode_heating/cathode_heating.py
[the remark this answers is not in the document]
bwalkerMIR: For what the purpose of reset_power_supply() is intended (as described in the function definition comments) and how it's used (it is run when the user requests 0 mA beam current so this function basically sets everything to zero make sure it really is off), I agree with Michael that the set current should remain at 0.

When setting a non-zero voltage, we should just make sure that the UCL has been set appropriately with enough head room to account for the max current we want during non-zero operation.

However, it looks like the code currently calculates the heater current and heater voltage for a given beam mA, and then tries to set the power supply to exactly those voltage and current settings. That will result in unpredictable behavior because if anything is slightly off in the real setup, you could either end up in voltage controlled mode or current controlled mode. We want to operate at all times in voltage controlled mode, with the exception that the power supply should have a cap on how much current can come out (provided by the OCP). As Joshua pointed out, we don't really know the exact resistance of our cathode and line, so a hypothetical "headroom" calculation may be difficult to implement in practice (although I agree a nice future improvement once our setup is better characterized). Honestly the simplest and most effective solution for now is likely to just set the voltage to whatever you need, and just set the current UCL to the same as OCP (or just below if that's problematic). That way we will know that we are operating in voltage controlled mode and still be protected from really bad over current events by the OCP.

subsystem/cathode_heating/cathode_heating.py
bwalkerMIR: I think we should keep the explicit current limit setting in there, but for now it should just be set equal to the OCP or just below. That should be good enough for cathode benchmarking. The headroom calculation is a little confusing about what it should be based on and set to. For example, we are going to get a new LUT with cathode benchmarking, so the headroom calc would in theory probably be based on that. But we don't have that yet. We also will do several LUT benchmarks, so each one may need to be recalculated. Also for benchtop testing with a dummy load, the V=IR load relationship may be different again, so given all the complexity I think we should avoid a dedicated headroom calculation for now.

theoreotically:
> I think we should keep the explicit current limit setting in there, but for now it should just be set equal to the OCP or just below. 

Done.

> The ES-440 V-I curves show that the resistance is not fixed. The resistance changes significantly with current and temperature.

Have restored changes in the code relevant to setting current, and have updated flowcharts to reflect the same.

>I think we should be able to change the voltage during operation without having to turn off the output first. If we change the voltage while the output is on, why don't we just have it also ramp up or down if the ramp is toggled on? If the ramp is toggled off, then it would be immediate and up to the user, but I still don't think we really need a warning.

Based on this comment, I have left in the changes that allow a user to change values during operation.

>why don't we just have it also ramp up or down if the ramp is toggled on?

Will double-check but Mark's branch (PR 62, which has already been approved) should have fixed our bug which prevented this from happening. The code as is supports ramping up & down, but I had identified a bug where we always set a voltage on output toggle before we decide on ramping or not, which meant that back when we had to toggle output on/off to change settings, in the event of setting 0.70 V and then changing to 0.68V our voltage curve went as follows:
0.00[output off] -> 0.70[output asserted] -> 0.00[ramp asserted] -> 0.70[ramp completed] -> 0.00[output off] -> 0.68[output asserted] -> 0.00[ramp asserted] -> 0.68[ramp completed]

theoreotically:
Relative ramping was always possible within the code, but now it should be able to actually go from a higher voltage value to a lower voltage value without starting from 0.

Still working on a fix for the power supply output voltage decreasing before it increases to set voltage when ramp is enabled AND output is just switched on. This bug does not present itself in other scenarios.

I removed the serial_lock, which seemed to be causing deadlock in certain scenarios. The most probable cause is the existence of deadlock, which led to a hang. Likely caused when a GETD (or similar) request interacting with the serial line takes the lock but encounters an unhandled exception or unhandled issue.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -43,32 +45,43 @@ def is_connected(self):
         return self.ser is not None and self.ser.is_open
 
     def flush_serial(self):
-        self.ser.reset_input_buffer()    
+        if self.ser and self.ser.is_open:
```
ColeMov, line 48: This appears to improve the flush_serial function by adding checks that a serial port is open prior to flushing.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -270,21 +319,24 @@ def get_voltage_current_mode(self):
...
+                            v2, c2, m2 = self.parse_getd_response(second_read)
+                            if v2 is not None and abs(v2) > 0.001:
+                                # overwrite with second read if it's nonzero
+                                self.log(f"Replaced 9104 0.0 V reading with second read {v2:.2f} V", LogLevel.VERBOSE)
+                                voltage, current, mode = v2, c2, m2
+                        return voltage, current, mode
+                self.log(f"Failed to get valid reading, attempt {attempt + 1}", LogLevel.WARNING)
+                time.sleep(0.1)
+            except Exception as e:
+                self.log(f"Error getting voltage mode", LogLevel.ERROR)
```
ColeMov, line 339: Implements exact same functionality but within a try loop for improved error handling.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1295,40 +1296,70 @@ def set_target_current(self, index, entry_field):
                 self.log(f"Interpolated heater current for Cathode {['A', 'B', 'C'][index]}: {heater_current:.3f}A", LogLevel.INFO)
                 self.log(f"Interpolated heater voltage for Cathode {['A', 'B', 'C'][index]}: {heater_voltage:.3f}V", LogLevel.INFO)
 
-                current_ovp = self.get_ovp(index)
-                if current_ovp is None:
-                    self.log(f"Unable to get current OVP for Cathode {['A', 'B', 'C'][index]}. Aborting voltage set.", LogLevel.ERROR)
-                    return
+                # set_voltage handles these checks now
```
sbasu107, line 1299: The checks are handled within the `set_voltage()` and `validate_voltage()` functions within `power_supply_94104.py`.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1441,49 +1469,61 @@ def update_predictions_from_voltage(self, index, voltage):
...
             if voltage > current_ovp:
-                self.log(f"Requested voltage ({voltage:.2f}V) exceeds OVP ({current_ovp:.2f}V) for Cathode {['A', 'B', 'C'][index]}. Aborting.", LogLevel.WARNING)
-                msgbox.showwarning("Voltage Exceeds OVP", f"The requested voltage ({voltage:.2f}V) exceeds the current OVP setting ({current_ovp:.2f}V). Please adjust the OVP or choose a lower voltage.")
-                return False
+                self.log(f"Calculated voltage ({voltage:.2f}V) exceeds OVP ({current_ovp:.2f}V) for Cathode {['A', 'B', 'C'][index]}. Aborting.", LogLevel.WARNING)
+                msgbox.showwarning("Voltage Exceeds OVP", f"The calculated voltage ({voltage:.2f}V) exceeds the current OVP setting ({current_ovp:.2f}V). Please adjust the OVP or choose a lower target current.")
+                return
+            
+            if voltage < 0:
+                raise ValueError("Voltage must be positive")
```
ColeMov, line 1483: Valid input checks for voltage.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1441,49 +1469,61 @@ def update_predictions_from_voltage(self, index, voltage):
...
+                        self.power_supplies[index].set_voltage(3, voltage)
+                        self.voltage_set[index] = True
                 
                 # Confirm the set values
-                set_voltage, set_current = self.power_supplies[index].get_settings(3)
-                if set_voltage is not None and set_current is not None:    
-                    voltage_mismatch = abs(set_voltage - voltage) > 0.01  # 0.01V tolerance
+                _ , set_current = self.power_supplies[index].get_settings(3)
+                if set_current is not None:    
+                    # voltage_mismatch = abs(set_voltage - voltage) > 0.01  # 0.01V tolerance
```
ColeMov, line 1515: This point to the end of code changes removes set_voltage variable

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -460,15 +513,34 @@ def configure_presets(self, setv1, seti1, swtime1, setv2, seti2, swtime2, setv3,
...
             self.log(f"{self.port} port already closed", LogLevel.INFO)
 
+    # def close(self):
+    #     """Close the serial connection."""
+    #     if self.ser and self.ser.is_open:
+    #         self.ser.close()
+    #         self.log(f"Closed serial port {self.port}", LogLevel.INFO)
+    #     else:
+    #         self.log(f"{self.port} port already closed", LogLevel.INFO)
+
```
ColeMov, line 543: Commented out copy of previous close function. I'd assume this can be deleted.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -460,15 +513,34 @@ def configure_presets(self, setv1, seti1, swtime1, setv2, seti2, swtime2, setv3,
...
+        self.log("Stopping threads and closing serial connection.", LogLevel.INFO)
+
+        # Signal threads to stop
+        self.stop_event.set()
+
+        # Wait for the ramping thread to finish
+        if self.ramp_thread and self.ramp_thread.is_alive():
+            self.ramp_thread.join()
+            self.log("Ramping thread terminated.", LogLevel.INFO)
+
```
ColeMov, line 528: Stops ramping threads before closing serial connection

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -150,12 +164,23 @@ def ramp_voltage(self, target_voltage, step_size=0.02, step_delay=2.0, preset=3,
...
+            daemon=True
         )
-        thread.start()
+        try:
+            self.ramp_thread.start()
+            self.log(f"Ramping voltage to {target_voltage:.2f}V started.", LogLevel.INFO)
+        except Exception as e:
+            self.log(f"Error starting ramping thread: {str(e)}", LogLevel.ERROR)
+            if callback:
+                callback(False)
```
sbasu107, line 183: Here there is additional error handling and logging added to the thread startup so that the thread behavior is easier to debug.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -176,16 +201,42 @@ def _ramp_voltage_thread(self, target_voltage, step_size, step_delay, preset, ca
...
+                            self.log(f"Attempt: {attempt} Failed to set voltage to {next_voltage:.2f}V.", LogLevel.ERROR)
+
+                    except Exception as e:
+                        self.log(f"Error during ramping step: {str(e)}. Aborting ramp.", LogLevel.ERROR)
+                        if callback:
+                            callback(False)
+                        return
                     
+                if attempt > self.MAX_RETRIES:
+                    self.log(f"Failed to set voltage to {next_voltage:.2f}V. Aborting ramp", LogLevel.ERROR)
```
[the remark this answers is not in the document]
ColeMov: An second conditional is required in order to check that the maximum number of attempts was reached as well as that the final attempt had failed.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1204,8 +1204,14 @@ def toggle_output(self, index):
...
                 return
             target_voltage = self.user_set_voltages[index]
+            
             if self.ramp_status[index]:
                 if target_voltage is not None:
+                    # Set voltage to 0 before starting the ramp-up
+                    self.power_supplies[index].set_voltage(3, 0.0)
+                    self.log(f"Voltage set to 0 for Cathode {['A', 'B', 'C'][index]} before ramping up.", LogLevel.DEBUG)
+                    
+                    # Ramp up to the target voltage
```
ColeMov, line 1214: Sets voltage on power supply to zero prior to ramping

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1295,40 +1296,70 @@ def set_target_current(self, index, entry_field):
...
+                #     self.log(f"Calculated voltage ({heater_voltage:.2f}V) exceeds OVP ({current_ovp:.2f}V) for Cathode {['A', 'B', 'C'][index]}. Aborting.", LogLevel.WARNING)
+                #     msgbox.showwarning("Voltage Exceeds OVP", f"The calculated voltage ({heater_voltage:.2f}V) exceeds the current OVP setting ({current_ovp:.2f}V). Please adjust the OVP or choose a lower target current.")
+                #     return
 
-                # Set Upper Voltage Limit and Upper Current Limit on the power supply
+                # Set Upper Current Limit on the power supply
                 if self.power_supplies and len(self.power_supplies) > index:
-                    self.log(f"Setting voltage: {heater_voltage:.2f}V and current: {heater_current:.2f}A", LogLevel.DEBUG)
-                    voltage_set_success = self.power_supplies[index].set_voltage(3, heater_voltage)
+                    self.log(f"Setting voltage: {heater_voltage:.2f}V", LogLevel.DEBUG)
```
ColeMov, line 1312: Sets voltage separate from current.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1295,40 +1296,70 @@ def set_target_current(self, index, entry_field):
...
+                    
+                    current_ovp = self.get_ovp(index)
+                    
+                    if current_ovp is None:
+                        self.log(f"Unable to get current OVP for Cathode {['A', 'B', 'C'][index]}. Aborting voltage set.", LogLevel.ERROR)
+                        return
+                    if heater_voltage > current_ovp:
+                        self.log(f"Calculated voltage ({heater_voltage:.2f}V) exceeds OVP ({current_ovp:.2f}V) for Cathode {['A', 'B', 'C'][index]}. Aborting.", LogLevel.WARNING)
+                        msgbox.showwarning("Voltage Exceeds OVP", f"The calculated voltage ({heater_voltage:.2f}V) exceeds the current OVP setting ({current_ovp:.2f}V). Please adjust the OVP or choose a lower target current.")
+                        return
```
ColeMov, line 1322: Checks that target voltage < OVP

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1441,49 +1469,61 @@ def update_predictions_from_voltage(self, index, voltage):
...
+            # Set Upper Current Limit on the power supply. Voltage is set in the power supply class when toggle_output is called
             if self.power_supplies and len(self.power_supplies) > index:
-                voltage_set_success = self.power_supplies[index].set_voltage(3, voltage)
+                # voltage_set_success = self.power_supplies[index].set_voltage(3, voltage)
                 current_set_success = self.power_supplies[index].set_current(3, heater_current)
-                if not voltage_set_success or not current_set_success:
-                    self.log(f"Unable to set voltage: {voltage} or current: {heater_current} for Cathode {['A', 'B', 'C'][index]}", LogLevel.ERROR)
-                    return False
+                if not current_set_success:
+                    self.log(f"Unable to set upper current limit: {heater_current} for Cathode {['A', 'B', 'C'][index]}", LogLevel.ERROR)
```
ColeMov, line 1496: Sets and checks current separate from voltage

instrumentctl/power_supply_9104/9104_supply_port_hub.md
```diff
@@ -1,5 +1,5 @@
...
-|--------|------|-----|
-|   a    |  4   |  4  |
-|   b    |  2   |  4  |
-|   c    |  3   |  4  |
\ No newline at end of file
+| supply | port |
+|--------|------|
+|   a    |  2   |
+|   b    |  3   |
+|   c    |  4   |
```
sbasu107, line 5: This Markdown file maps the COM port numbers to the appropriate power supply for each cathode.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -43,32 +45,43 @@ def is_connected(self):
...
-                return response.strip()
-            except serial.SerialException as e:
-                self.log(f"Serial error: {e}", LogLevel.ERROR)
-                return None
-            except ValueError as e:
-                self.log(f"Error processing response for command '{command}': {str(e)}", LogLevel.ERROR)
-                return None
+            self.log(f"Response: {response}", LogLevel.DEBUG)
+                
+            return response.strip()
```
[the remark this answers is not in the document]
ColeMov: This ideally prevents deadlocks/hangs when sending serial commands.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -43,32 +74,89 @@ def is_connected(self):
...
+            "event": threading.Event(),
+            "response": None,
+        }
+        self._cmd_queue.put(job)
+
+        # Wait (with timeout slightly longer than port timeout)
+        if not job["event"].wait(self.timeout + 1.0):
+            self.log(f"Timeout waiting for reply to '{command}'", LogLevel.ERROR)
+            return None
+        return job["response"]
```
ColeMov, line 113: Command is queued and operation holds until response is read. The exact timing of this may need to be tweaked.

instrumentctl/power_supply_9104/power_supply_9104.py
```diff
@@ -460,13 +605,30 @@ def configure_presets(self, setv1, seti1, swtime1, setv2, seti2, swtime2, setv3,
...
+            self.ramp_thread.join(timeout=2)
+            self.log("Ramping thread terminated.", LogLevel.INFO)
+
+        # Terminate serial worker and empty queue
+        if self.serial_thread and self.serial_thread.is_alive():
+            self._cmd_queue.put(None)
+            self.serial_thread.join(timeout=2)
+
+
+        # Close the serial connection
```
sbasu107, line 627: Added more proper handling of stopping both the ramping and serial worker threads before closing the serial port. This helps to make shutdowns safer and prevent possible issues from background activity.

PR #59, opened 2025-03-13
Bugfix/save layout

AryanSV5:
Ensured the save layout function works. We previously didn't save height and width of the Messages Frame within frames_config in dashboard.py. I made sure these values were saved upon clicking the save layout button and resized some of the child widgets within the Messages Frame to ensure they fit according to the given height and width values. 

Ran a test on 3/13/2025 at 5:20pm with all subsystems hooked up and the save layout button is confirmed to work. I simply adjusted the panels and restarted the application to make sure the new panel dimensions were reflected upon the restart. 

save_layout_test_3_13_2025


mslaffin (approved):
this is great! and I can confirm it appears to work well on my machine so I'm going to merge

PR #60, opened 2025-03-15
Feature/ccs button toggle (Ramp toggle)

mark11778:
image

Current status ^

Added Feature to allow toggle to either set voltage or ramp to set voltage, red indicating that there is no ramp, green that it will ramp.

let me know if there is any edits regarding functionality or style that should be changed


mslaffin:
Hey Mark, the ramp status tracking that you implemented makes sense to me, but I am thinking we can still improve the clarity of this. I have some ideas

The color red might be inappropriate because of its association with problem states.

To me, directly setting the voltage (opposed to using ramp_mode) shouldn't really be construed as a machine error because it's just an operating configuration with slightly higher risk. 

Encoding the ramp mode information in color can obfuscate this. 

We could use radio buttons instead, and default to the "safer" mode (ramping):
image
This does however reveal that we need to resize the temperature plots with the parent frame.
@bwalkerMIR do you have any thoughts?

We may also want an "active" indicator to show if the ramp is in-progress. like:
image

mslaffin:
> Hey Michael, did you implement what is in the photo or where is that coming from?

Yea, I did actually implement this. I didn't want to overwrite your work immediately, but I'll push these changes here so that you and @bwalkerMIR can review. If we decide against it, we can always revert back to your version.
<br>
> if we are using the non-ramp set something is not in the right state overall of the machine. Thus, it could justify some higher level error state.

It's not clear to me why this is assumed here... I think it's possible we might determine that a 0.1V/s ramp is just overly conservative at some later point. Low set voltages, thermal mass of the cathode support structure, 9104 voltage slew rate, cable inductances, etc. might naturally limit the power delivery (and thermal shock) to the cathode. 

LaB6 has a relatively low coefficient of thermal expansion, and this ES-423E series [note from Kimball Physics ](https://www.emsdiasum.com/docs/technical/manuals/81080.pdf?srsltid=AfmBOorgJMcTjmFlC4DH3QfPr3FaWvxyHilf6aA9IZMAeKIMQKE1jJsq) even states that after an initial run-up, "the cathode can be turned off and on to maximum current instantly". Whether this is applicable to the ES-440 series remains unclear to me. 

At any rate, the radio buttons present both options as legitimate choices, rather than suggesting one is "wrong".

bwalkerMIR:
I like the green status LED that shows if it's currently ramping, I also like the radio buttons better as it's more clear what's happening and agree the red coloring for one is not necessarily the same level as an error. The gui got a bit wider so I would rearrange a bit, CCS is already too wide on any 1080p screens.

Wow that ES423 operations link is really great! Yes we might be being overly cautious with this, but will still try to start slow for now.

One of my questions was whether this toggle affects both if you use the set beam current or if you use the set voltage. gpto1-pro and grok3 both agree this is true and uses the same ramp functionality, which is what I think it should do.
image

bwalkerMIR (approved):
See my last comment in the PR thread. Mark and I walked through the files changes and the thread with Michael and Mark. This looks great and functionality makes sense.

PR #62, opened 2025-03-18
Feature/oil like vacuum

arundhatisingh17:
Brandon asked for a different GUI Layout. I have uploaded Brandon's annotations and the current state of the GUI here for reference. 

image (1)

Screenshot 2025-03-18 at 4 28 56 PM

The number of rows have been reduced from 6 to 5. All the components have been shifted accordingly. Instead of creating a separate frame for setup scripts, I have placed in in the main tab of the "Main Control" Panel.


bwalkerMIR:
@arundhatisingh17 Before this PR is reviewed, this asana task should be completed: [Merge develop into "oil like vacuum" branch and update PR so that the feature branch is up to date with most recent changes in develop](https://app.asana.com/0/1208196184307227/1209728232533217/f). 

This PR has GUI changes but the branch is not up to date with develop. This feature branch should have develop merged into it to simplify the PR review and make sure that the PR is really just new changes. I suspect that since Aryan already merged many gui changes that there is likely some conflict or redundancy between PR 62 and the up to date develop branch which includes Aryan's work.

This task is to perform the following, merging the most recent develop changes into the "oil like vacuum" branch and then updating the feature branch PR so there are fewer changes and they are consistent

.gitignore
bwalkerMIR, line 1: The .gitignore needs to have usr/usr_data added back in before being merged to develop

main.py
```diff
@@ -17,6 +17,36 @@
     'ProcessMonitors'
 ]
 
+"""Dictionary storing all open serial connections"""
```
mark11778, line 20: I am a bit confused why these are in main.py? What was the thought process here?

I have a feeling that if you start these Serial objects in here, then you won't be able to start the serial ports later in when the subsystems are initialized in dashboard because the ports will be "in use" by these serial objects.

subsystem/interlocks/interlocks.py
```diff
@@ -54,7 +54,11 @@ def __init__(self, parent, com_ports, logger=None, frames=None, active=None):
             if com_ports is not None:
                 try:
                     self.driver = g9_driv.G9Driver(com_ports, logger=self.logger)
-                    self.log("G9 driver initialized", LogLevel.INFO)
+                    if self.driver.wait_for_connection(timeout = 30):
```
mark11778, line 57: Why where these edits made?

bwalkerMIR (changes requested):
There are several changes that should be made prior to finishing the review on this PR because the PR is overcomplicated by a bunch of changes that are unnecessary.

1. There isn't actually any change to .gitignore, get rid of the non-functional change
2. dashboard.py also has no functional changes, get rid of them so it shows nothing needs to be reviewed here,
3. There is a comment block in g9_driver.py that shouldn't be committed to develop
4. There is a big comment block in main.py that shouldn't be committed to develop
5. The line removal from cathode_heating.py should probably be reverted
6. There is some debate about the outer try statement in interlocks.py. I'm curious what @mslaffin thinks here
7. oil_system.py has been changed to reflect a simpler version that is smaller and looks great. I haven't reviewed in detail the code in here yet but it looks good in the gui.
8. The non-functional line removal in vtrx.py should be removed unless this is really wanted. Please confirm.
9. com_ports.json should not be committed to develop. This is in the gitignore
10. pane_state.json should not be committed to develop. This is in the gitignore

subsystem/interlocks/interlocks.py
bwalkerMIR: I am wondering the same thing

subsystem/interlocks/interlocks.py
```diff
@@ -50,22 +50,16 @@ def __init__(self, parent, com_ports, logger=None, frames=None, active=None):
...
-            else:
-                self.driver = None
-                self.log("No COM port provided for G9 driver", LogLevel.WARNING)
-                self._set_all_indicators('red')
-        except Exception as e:
-            self.driver = None
-            self.log(f"Failed to initialize G9 driver: {str(e)}", LogLevel.WARNING)
-            self._set_all_indicators('red')
+	
+
```
mslaffin, line 62: > There is some debate about the outer try statement in interlocks.py. I'm curious what @mslaffin thinks here

I'm thinking this change is unsubstantiated and should be reverted. 

The original outer try block provides a greater safety net to catch errors that might occur outside the driver creation. You might have found this unnecessary in testing, but it's added as a defensive pattern that incurs no cost and should be adopted widely in our code. 

And as BW has noted, we should definitely revert back to the original else clause as well.

subsystem/oil_system/oil_system.py
```diff
@@ -1,350 +1,93 @@
-# oil_system.py
 import tkinter as tk
-from tkinter import font as tkFont
-import math
+from tkinter import ttk, messagebox
```
bwalkerMIR, line 2: ttk, messagebox, datetime are never used → delete imports if not used.

subsystem/oil_system/oil_system.py
```diff
@@ -1,350 +1,93 @@
...
-            text=name, 
-            font=('Arial', 8, 'bold'), 
-            anchor='w'
-        )
+class OilSubsystem:
+    def __init__(self, parent, logger=None):
+        """Initializes the OilSubsystem inside the given parent (Frame or Main Window)."""
+        self.parent = parent
+        self.logger = logger
+        self.pressure = 3.5
```
bwalkerMIR, line 13: pressure should probably be initialized with 0, temperature with 20, flow rate at 0 again.

I don't think we should be initializing these variables with values in the middle of the range when it's not real, but if I'm missing something please correct me.

subsystem/oil_system/oil_system.py
```diff
@@ -1,350 +1,93 @@
...
+class OilSubsystem:
+    def __init__(self, parent, logger=None):
+        """Initializes the OilSubsystem inside the given parent (Frame or Main Window)."""
+        self.parent = parent
+        self.logger = logger
+        self.pressure = 3.5  
+        self.temperature = 50.0  
+        self.flow_rate = 10.0  
+        self.pump_status = False  
+        self.stop_event = threading.Event()
```
bwalkerMIR, line 17: The simulated data is cool and use of its own thread is good coding. However, since we are going to use this gui very shortly for real experiments, I don't think we should be merging in simulated data. I think the simulated data and extra thread should be removed from this code. The oil system should just display that they aren't getting data in a way that is similar to how we show that in other subpanels in the gui.

subsystem/oil_system/oil_system.py
```diff
@@ -1,350 +1,93 @@
...
-        self.create_text(
-            self.center_x, 
-            self.height - 45,
-            text="--.- PSI",
-            font=('Helvetica', 12, 'bold'),
-            tags="value_text"
-        )
+        self.create_sensor_frame("Temperature", f"{self.temperature:.1f}°C", "temp_label")
+        self.create_sensor_frame("Pressure", f"{self.pressure:.1f} PSI", "pressure_label")
+        self.create_sensor_frame("Flow Rate", f"{self.flow_rate:.1f} L/min", "flow_label")
```
bwalkerMIR, line 32: flow rate is created as "L/min" but we display in the gui as "GPM". This is confusing. Let's just stick with GPM everywhere to be consistent.

subsystem/oil_system/oil_system.py
bwalkerMIR: also note we probably don't need threading if we get rid of the simulated data

bwalkerMIR (approved):
This code looks good to go and creates a simplified oil system subpanel that can be integrated later with electronics once built.

PR #63, opened 2025-03-19
Feature/misleading initialization log message

arundhatisingh17:
I commented out lines that used a helper method which would further block the main threads from updating the GUI elements. The tkinter library expects the main thread (instead of the background threads) to update the tkinter elements.

The section in interlocks.py where we check for connection is enclosed in a try-except blocks where and the Warningmessage is only printed out when connection fails.


bwalkerMIR:
I'm going to wait until PR 62 is merged before finishing my review of this PR 63, because this PR 63 includes many of the changes that are already in PR 62. Once PR 62 is merged then we can see how different PR 63 really is. @arundhatisingh17

bwalkerMIR:
Closing this PR because current implementation does not function as intended and it's simpler to restart the changes from a fresh branch off the latest develop. There are only 5 lines of changes in a single file (interlocks.py) and they don't work right (results in an exception). 
image

This PR was branched off from develop a long time ago so shows 51 commits for only 5 lines of edits. We should delete the branch, close the PR and restart fresh, considering it doesn't work. 

The reason it doesn't work is that the new edit calls `self.driver.wait_for_connection`, but such a method does not exist in the whole repo, resulting in an exception for the code that was added so it totally gets bypassed anyways. The method should have been added to g9_driver.py but wasn't so nonfunctional.

PR #64, opened 2025-04-23
Feature/lookup table

mark11778:
Important thing to note:
New dependency in requirements.txt, thus after this merge will need to pip install -r requiements.txt

There are 3 files for the look up tables(currently living in subsystem/cathode_heating), each one representing a power supply/cathode pair. right now they do not reflect any data that is different from the interpolate functions that Michael has already created. If there is a better place for these files to live I am happy to move them.

original:
```mermaid
flowchart TD
 subgraph s1["update_predictions_from_voltage"]
        n1["sanitize inputs "]
        n2["interpolate"]
        n3["updates set voltage and current"]
  end
    A["on_voltage_label_click"]


n1 --> n2 --> n3

A --> s1

```

new implementation:
```mermaid
flowchart TD
 subgraph s1["update_predictions_from_voltage"]
        n1["sanitize inputs "]
    subgraph s2["cur_vlt_converter"]
            m1["user selected generator"]
            m2["look up table"]
            m3["interpolate"]
            m4["return val"]
    end
    n3["updates set voltage and current"]
  end
    A["on_voltage_label_click"]


n1 --> s2 

m1 --> m2
m1 -->m3 
m3 --> m4
m2 --> m4

m4--> n3

A --> s1
```

the implement for converting the from the mA current is pretty much the same except it returns both the voltage and the current value off of the given mA.


bwalkerMIR:
Just putting a reminder for myself here when I look back but this PR is not ready yet and we are focusing on an MVP for cathode benchmarking instead right now since higher priority

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1594,4 +1633,82 @@ def close_com_ports(self):
...
+            
+        Returns:
+            float: Interpolated value
+        """
+        if vltToCur:
+            return self.cur_cathode_model.interpolate(val, inverse=True)
+        else:
+            heater_current = self.emission_current_model.interpolate(val, inverse=True)
+            heater_voltage = self.heater_voltage_model.interpolate(heater_current)
+            return heater_voltage
```
ColeMov, line 1714: Does this replace the current interpolate function in ES440_cathode?

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1450,16 +1479,27 @@ def update_predictions_from_voltage(self, index, voltage):
...
+                        return False
+                    voltage = new_voltage
 
+            # self.log(f"{emission_voltage=} : {heater_current=} : {predicted_current=} ___________________", LogLevel.ERROR)
 
             # Check if the interpolated current is within the model's range
-            if not min(cathode_model.x_data) <= heater_current <= max(cathode_model.x_data):
-                self.log(f"Heater current {heater_current:.3f} is out of range [{min(cathode_model.x_data):.3f}, {max(cathode_model.x_data):.3f}]", LogLevel.WARNING)
+            if not min(self.cur_cathode_model.x_data) <= heater_current <= max(self.cur_cathode_model.x_data):
+                self.log(f"Heater current {heater_current:.3f} is out of range [{min(self.cur_cathode_model.x_data):.3f}, {max(self.cur_cathode_model.x_data):.3f}]", LogLevel.WARNING)
```
ColeMov, line 1502: Cathode model changes based on config dropdown.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -78,7 +82,18 @@ def __init__(self, parent, com_ports, active, logger=None):
...
+        self.current_options = {
+            "Cathode A" : pd.read_csv('subsystem/cathode_heating/powersupply_A.csv'),
+            "Cathode B" : pd.read_csv('subsystem/cathode_heating/powersupply_B.csv'),
+            "Cathode C" : pd.read_csv('subsystem/cathode_heating/powersupply_C.csv'),
+            "Interpolate" : self.interpolate
+        }
+        self.interpolate_setting = [self.current_options["Cathode A"], 
+                                    self.current_options["Cathode B"], 
+                                    self.current_options["Cathode C"],
+                                    self.current_options["Interpolate"]]
```
sbasu107, line 94: Letting users select interpolation sources per cathode

sbasu107:
Today, I discussed with @bwalkerMIR and @ColeMov about removing `interpolate` and relying on the Cathode Benchmarking data. I started commenting out the relevant sections and am currently testing along the way to ensure that everything else still functions as expected.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -453,6 +469,18 @@ def setup_gui(self):
                                         command=lambda unit=i+1: self.read_and_log_temperature(unit))
             read_temp_button.grid(row=11, column=0, columnspan=2, sticky='ew', padx=5, pady=2)
 
+            # Add dropdown for interpolate_setting
+            interpolate_label = ttk.Label(config_tab, text='Select Lookup Table:', style='RightAlign.TLabel')
```
sbasu107, line 473: Changed 'Select Interpolate Setting' to 'Select Lookup Table'.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -254,52 +258,36 @@ def setup_gui(self):
...
-            ttk.Label(main_tab, textvariable=self.predicted_grid_current_vars[i], style='Bold.TLabel').grid(row=3, column=1, sticky='w')
+            ttk.Label(main_tab, textvariable=self.predicted_grid_current_vars[i], style='Bold.TLabel').grid(row=2, column=1, sticky='w')
             
             # Predicted heater current (A)
-            ttk.Label(main_tab, text='Pred Heater Current (A):', style='RightAlign.TLabel').grid(row=4, column=0, sticky='e')
-            ttk.Label(main_tab, textvariable=self.predicted_heater_current_vars[i], style='Bold.TLabel').grid(row=4, column=1, sticky='w')
-
-            # Predicted cathode temperature (C)
-            ttk.Label(main_tab, text='Pred CathTemp (C):', style='RightAlign.TLabel').grid(row=5, column=0, sticky='e')
-            ttk.Label(main_tab, textvariable=self.predicted_temperature_vars[i], style='Bold.TLabel').grid(row=5, column=1, sticky='w')
```
sbasu107, line 295: Deleted line for Predicted cathode temperature

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -337,21 +325,21 @@ def setup_gui(self):
...
             
             # Actual target current (mA)
-            ttk.Label(main_tab, text='Actual Target (mA):', style='RightAlign.TLabel').grid(row=9, column=0, sticky='e')
-            ttk.Label(main_tab, textvariable=self.actual_target_current_vars[i], style='Bold.TLabel').grid(row=9, column=1, sticky='w')
+            ttk.Label(main_tab, text='Actual Target (mA):', style='RightAlign.TLabel').grid(row=7, column=0, sticky='e')
+            ttk.Label(main_tab, textvariable=self.actual_target_current_vars[i], style='Bold.TLabel').grid(row=7, column=1, sticky='w')
             
             # Temperature monitoring (C)
-            ttk.Label(main_tab, text='Actual ClampTemp (C):', style='RightAlign.TLabel').grid(row=10, column=0, sticky='e')
+            ttk.Label(main_tab, text='Actual ClampTemp (C):', style='RightAlign.TLabel').grid(row=8, column=0, sticky='e')
```
sbasu107, line 340: These values in the main menu had to be shifted up as two lines were deleted

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -417,41 +405,57 @@ def setup_gui(self):
...
+            lookup_table_label.grid(row=5, column=0, sticky='e')
+
+            lookup_table_options = list(self.current_options.keys())
+            lookup_table_box = ttk.Combobox(config_tab, values=lookup_table_options, state='readonly')
+            lookup_table_box.grid(row=5, column=1, sticky='w')
+
+            lookup_table_box.set(f"Cathode {['A', 'B', 'C'][i]}")
+            lookup_table_box.bind("<<ComboboxSelected>>", lambda event, idx=i: self.on_lookup_table_change(event, idx))
+
+            self.lookup_table_comboboxes.append(lookup_table_box)
```
sbasu107, line 419: Instead of interpolate settings, this dropdown menu is only for the look up tables.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1236,153 +1235,6 @@ def toggle_output(self, index):
...
-                        return
-                    if heater_voltage > current_ovp:
-                        self.log(f"Calculated voltage ({heater_voltage:.2f}V) exceeds OVP ({current_ovp:.2f}V) for Cathode {['A', 'B', 'C'][index]}. Aborting.", LogLevel.WARNING)
-                        msgbox.showwarning("Voltage Exceeds OVP", f"The calculated voltage ({heater_voltage:.2f}V) exceeds the current OVP setting ({current_ovp:.2f}V). Please adjust the OVP or choose a lower target current.")
-                        return
-                    # Set the upper current limit on the power supply
-                    
-                    # voltage_set_success = self.power_supplies[index].set_voltage(3, Decimal(heater_voltage))
-                    current_set_success = self.power_supplies[index].set_current(3, heater_current)
-
```
sbasu107, line 1327: We already handle setting Overcurrent limit on the power supply in set_overcurrent_limit

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1236,153 +1235,6 @@ def toggle_output(self, index):
...
-    def set_target_current(self, index, entry_field):
-        """
-        Set target beam current for a cathode and calculate required heater settings.
-
-        Uses the target beam current to calculate ideal emission current, then determines
-        the appropritate heater voltage and current using the ES440 cathode data model.
-
-        Args:
-            index (int): Index of the cathode (0-2)
-            entry_field (ttk.Entry): Entry widget containing target current value
```
sbasu107, line 1249: This function is redundant as we no longer are setting a target current from the dashboard

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1482,13 +1332,55 @@ def update_predictions_from_voltage(self, index, voltage):
...
+            while True:
+                try:
+                    heater_voltage, heater_current, target_current = self.emission_cur_vlt_converter(index, voltage)
+                    
+                    # Check if lookup table returned zero values (voltage out of range)
+                    if heater_current == -1 and target_current == -1:
+                        # Still allow voltage setting, but don't update predictions
+                        self.predicted_heater_current_vars[index].set('--')
+                        self.predicted_emission_current_vars[index].set('--')
+                        self.predicted_grid_current_vars[index].set('--')
```
sbasu107, line 1344: See my comment at the end of this file under the emission_cur_vlt_converter() function. If we are not able to finds an exact match in the lookup table, we return -1 for the values we needed and keep showing dashes (--) for those values.

sbasu107:
image
image

To demonstrate that only values that are in the lookup table will populate the main menu, I am showing in the top image what happens if you look up 0.66V, which is in the table, and 0.67 V, which is not in the table.  Datasets A, B, and C for the lookup tables are the exact same right now so the lower image applies to all of them.

subsystem/cathode_heating/cathode_heating.py
```diff
@@ -1631,7 +1523,52 @@ def close_com_ports(self):
...
+        """
+        if isinstance(self.lookup_table_setting[index], pd.DataFrame):
+            df = self.lookup_table_setting[index]
+            # Look for exact match in voltage column
+            exact_match = df[df['voltage'] == val]
+            if exact_match.empty:
+                heater_voltage = val
+                heater_current = -1
+                target_current = -1
+                return (heater_voltage, heater_current, target_current)
```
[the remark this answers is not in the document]
sbasu107: We don't care about intermediate values anymore

bwalkerMIR:
> To demonstrate that only values that are in the lookup table will populate the main menu, I am showing in the left image what happens if you look up 0.66V, which is in the table, and 0.67 V, which is not in the table. The test datasets A, B, and C for the lookup tables are the exact same right now so the right image applies to all of them.


Why are we allowing the user to enter 0.67 V in the first place if the power supply can only go up in 0.02 V increments? Why not just require the user to set in 0.02 V increments so that with most of our lookup table all the values populate properly rather than dashes?

bwalkerMIR:
> I would be happy to make those changes, but may I do them in the feature/CCS-current-control branch that @ColeMov was working on? That branch already has the lookup functionality from this branch and +0.02 V/−0.02 V buttons from the Cathode Heating main menu redesign, so I think it may be a better place to test appropriate value setting. I wanted to ensure that this branch is just for testing value lookup, while the menu layout and related features are being handled in feature/CCS-current-control.

Ah yes this makes sense. Let's not complicate this PR. What currently happens though if you enter a value that isn't in the lookup table?

> As a final thought, would it make sense to either show an error popup when the user enters an unsupported value, or automatically round it (up or down) to the nearest valid increment?

Hmm I think it would be best to not allow the user to set such a value in the first place. Rounding would be a bit of a hidden feature and I think could get confusing. I guess an error poppup probably makes the most sense.

sbasu107:
> > I would be happy to make those changes, but may I do them in the feature/CCS-current-control branch that @ColeMov was working on? That branch already has the lookup functionality from this branch and +0.02 V/−0.02 V buttons from the Cathode Heating main menu redesign, so I think it may be a better place to test appropriate value setting. I wanted to ensure that this branch is just for testing value lookup, while the menu layout and related features are being handled in feature/CCS-current-control.
> 
> Ah yes this makes sense. Let's not complicate this PR. What currently happens though if you enter a value that isn't in the lookup table?
> 
> > As a final thought, would it make sense to either show an error popup when the user enters an unsupported value, or automatically round it (up or down) to the nearest valid increment?
> 
> Hmm I think it would be best to not allow the user to set such a value in the first place. Rounding would be a bit of a hidden feature and I think could get confusing. I guess an error popup probably makes the most sense.

If you enter a value not present in the lookup table, the predicted values won’t populate and will instead appear as dashes. However, as long as your requested voltage and/or current values remain below the overvoltage and overcurrent limits, you can still set that value successfully on the power supply.

Here are pictures of this happening on the e-beam laptop:

<table align="center">
  <tr>
    <td align="center">
      feature/lookup-table
      <br>
      <em>feature/lookup-table</em>
    </td>
    <td align="center">
      feature/CCS-current-control
      <br>
      <em>feature/CCS-current-control</em>
    </td>
  </tr>
</table>

PR #65, opened 2025-04-26
Bugfix/sic

mark11778:
during testing today it was found that sic was not working correctly. i found a bug that was causing it on the lab laptop which was under some circumstances, the sic g9driver would attempt to output to the log, the log accessing the tkinter main loop, which tkinter does not like it when separate threads are accessing it. so i refactored the driver and backend gui to instead of writing to the log, the driver raises and error - when appropriate - which is then parse in the interlocks.py and then outputted to the log.

also found a bug in how the unit status was being parse as we were not consider the reserved bits values in the previous implementation, which now we are ignoring the reserve and only returning the flag bits. 

currently able to open all (both on my computer and lab laptop like below  \/)
image

### DIFF
--- a/instrumentctl/G9SP_interlock/g9_driver.py
+++ b/instrumentctl/G9SP_interlock/g9_driver.py
@@ -3,7 +3,6 @@ import serial
 import threading
 import queue
 import time
-from utils import LogLevel
 
 class G9Driver:
     NUMIN = 13
@@ -406,13 +405,6 @@ class G9Driver:
             extracted_bits.extend(((byte >> i) & 1) for i in range(bits_to_extract - 1, -1, -1)[::-1])
 
         return extracted_bits[:num_bits]
-
-    def log(self, message, level=LogLevel.INFO):
-        """Log a message with the specified level if a logger is configured."""
-        if self.logger:
-            self.logger.log(message, level)
-        elif self.debug_mode:
-            print(f"{level.name}: {message}")
 
 
     #TODO: Figure out how to handle all the errors (end task)
--- a/instrumentctl/G9SP_interlock/g9_driver.py
+++ b/instrumentctl/G9SP_interlock/g9_driver.py
@@ -58,7 +58,6 @@ class G9Driver:
     }
 
     def __init__(self, port=None, baudrate=9600, timeout=0.5, logger=None, debug_mode=False):
-        self.logger = logger
         self.debug_mode = debug_mode
         self.ser = None
         self.setup_serial(port, baudrate, timeout)
--- a/instrumentctl/G9SP_interlock/g9_driver.py
+++ b/instrumentctl/G9SP_interlock/g9_driver.py
@@ -78,7 +78,6 @@ class G9Driver:
         """
         if port:
             try:
-                
                 self.ser = serial.Serial(
                     port=port,
                     baudrate=baudrate,
@@ -86,34 +85,24 @@ class G9Driver:
                     stopbits=serial.STOPBITS_ONE,
                     bytesize=serial.EIGHTBITS,
                     timeout=timeout
-                    )
-                            
-                self.log(f"Serial connection established on {port}", LogLevel.INFO)
-            except serial.SerialException as e:
+                    )   
+            except serial.SerialException:
                 self._close_serial()
-                self.log(f"Failed to open serial port {port}: {str(e)}", LogLevel.ERROR)
         else:
             self._close_serial()
-            self.log("No port specified", LogLevel.WARNING)
+            raise ConnectionError
 
     def _close_serial(self):
         """ Attempt to close serial port """
         if self.ser and self.ser.is_open:
             self.ser.close()
-            self.log("Serial Connection was closed for G9", LogLevel.INFO)
-        else:
-            self.log("Serial connection is already close", LogLevel.DEBUG)
         self.ser = None
 
     def _update_queue(self, response=None):
-        try:
-            data = response if response else ([0] * 13, [0] * 13, 0)
-            if self._response_queue.full():
-                self._response_queue.get_nowait()
-            self._response_queue.put(data)
-        except Exception as e:
-            self.log(f"Error storing response in queue {data=}: {e=}")
-
+        data = response if response else ([0] * 13, [0] * 13, 0)
+        if self._response_queue.full():
+            self._response_queue.get_nowait()
+        self._response_queue.put(data)
 
     def _communication_thread(self):
         """Background thread for handling serial communication"""
@@ -129,18 +118,13 @@ class G9Driver:
                     if response_data:
                         result = self._process_response(response_data)
                         self._update_queue(result)
-
-            except (ValueError, TimeoutError) as e:
-                self.log(f"Error was throw while getting package from G9: {e}", LogLevel.CRITICAL)
            
-            except PermissionError as e:
-                self.log(f"PermissionError while reading from serial port: {str(e)}", LogLevel.ERROR)
+            except PermissionError:
                 self._update_queue()
                 self._running = False
                 self._close_serial()
 
-            except serial.SerialException as e:
-                self.log(f"Communication thread error: {str(e)}", LogLevel.ERROR)
+            except serial.SerialException:
                 self._update_queue()
 
             time.sleep(0.1)  # minimum sleep between successful reads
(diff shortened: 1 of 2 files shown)


mark11778:
the unit tests will need to be updated so it does not fail the pipeline, but i can work on that tomorrow

bwalkerMIR:
Nice!! Are you in the lab?! I'll take a look at the file changes too

mark11778:
yeah i was, but just got home

theoreotically:
Glad you were able to fix that so soon. I think we need to eventually come up with documentation that isn't the 50-page dev specs. It should allow fresh/future members to understand what is going on in each subsystem even if they haven't worked on it from crash for issues just like today's. 

The logger mechanism may need an overhaul to fix some of our issues. Our patches involve consolidating code into higher-level systems/subsystems for now as opposed to each driver.  It is a bit like throwing drain cleaner down a pipe, instead of addressing the root causes. 

If possible, a future sprint may need to involve addressing just GUI problems. It would need to be pushed back quite a bit but can be reworked based on how summer testing goes.

citestfiles/G9Driver/g9driver_test.py
```diff
@@ -34,7 +34,8 @@ class TestG9Driver(unittest.TestCase):
 
     def setUp(self):
         self.mock_serial = MagicMock()
-        self.driver = G9Driver()
+        self.mock_serial.port = "COM1"  # Add port name to mock_serial
+        self.driver = G9Driver(port="COM1")  # Pass port to G9Driver
```
sbasu107, line 38: Modified test setup to explicitly set the mock serial port to "COM1" while initializing G9Driver for more realistic testing.

citestfiles/G9Driver/g9driver_test.py
```diff
@@ -54,39 +55,14 @@ def test_normal_response_processing(self):
         self.driver.ser.read_until.return_value = msg_with_checksum
         
         try:
-            sitsf, sitdf, g9Active = self.driver._process_response(msg_with_checksum)
+            sitsf, sitdf, g9Active, _ , _, _ = self.driver._process_response(msg_with_checksum)
```
sbasu107, line 58: We update test_normal_response_processing to handle additional return values. 

This because the _process_response method in `g9driver` was updated to also return `unit_status`, as well as error codes for the safety input and output terminals.

citestfiles/G9Driver/g9driver_test.py
```diff
@@ -206,8 +135,10 @@ def test_comport_connection_none(self):
         """Testing with passing none to the method"""
         self.driver.ser = serial.Serial(port=None)
         self.assertIsNotNone(self.driver.ser)
-        self.driver.setup_serial(None)
-        self.assertIsNone(self.driver.ser)
+        try:
+            self.driver.setup_serial(None)
+        except ConnectionError:
+            self.assertIsNone(self.driver.ser)
```
sbasu107, line 141: When no port is specified, we expect `test_comport_connection_none()` to return a value to raise a `ConnectionError`.

instrumentctl/G9SP_interlock/g9_driver.py
```diff
@@ -3,7 +3,6 @@
 import threading
 import queue
 import time
-from utils import LogLevel
```
sbasu107, line 6: Removed direct logging from G9Driver, since there were Tkinter threading issues during testing.

instrumentctl/G9SP_interlock/g9_driver.py
```diff
@@ -153,12 +136,13 @@ def get_interlock_status(self):
...
-            # peep the queue without removing the status
+            # Try to get an item from the queue without removing it
             item = self._response_queue.get_nowait()
-            self._response_queue.put(item) # put it back
+            # Put it back since we just wanted to peek
+            self._response_queue.put(item)
             return item
         except queue.Empty:
-            self.log("No interlock information is here; Queue is Empty", LogLevel.WARNING)
+            # Queue is empty, return None
```
sbasu107, line 145: For getting the safety status of the controller's input and output terminals, logging was removed and now for an empty queue we simply return none.

instrumentctl/G9SP_interlock/g9_driver.py
```diff
@@ -250,15 +223,16 @@ def _process_response(self, data):
...
+
+        unit_flags = {self.US_STATUS[k] : unit_status_flags[k] for k in self.US_STATUS.keys()}
 
-        return (binary_data['sitsf'], binary_data['sitdf'],
-                 binary_data['sotsf'][4] & binary_data['sotdf'][4])
+        return (binary_data['sitsf'], binary_data['sitdf'],                 # sitsf_bits , sitdf_bits
+                    binary_data['sotsf'][4] & binary_data['sotdf'][4],      # g9_active
+                    unit_flags,                                             # unit_status
+                    data[self.SITEC_OFFSET:self.SITEC_OFFSET + 24][-10:],   # input 
+                    data[self.SOTEC_OFFSET:self.SOTEC_OFFSET + 16][-10:])   # output
```
sbasu107, line 235: Removed logging and error checks for thread safety and added more detailed data, including unit status flags and as well as error codes for the safety input and output terminals.

instrumentctl/G9SP_interlock/g9_driver.py
```diff
@@ -315,65 +289,6 @@ def _validate_checksum(self, data):
                 f" Received: {received.hex()}"
             )
 
-    def _check_unit_status(self, status):
-        """
-        Check unit status and raise error if issues found
-
-        Raise:
```
sbasu107, line 322: `_check_unit_status`, '_check_safety_inputs', 'check_safety_outputs', `_check_terminal_status` removed to prevent thread-safety issues with Tkinter.

subsystem/interlocks/interlocks.py
```diff
@@ -232,8 +271,29 @@ def update_data(self):
...
+                sitsf_bits, sitdf_bits, g9_active, unit_status, input_terms, output_terms = status
+
+                # parse unit status
+                for k, v in unit_status.items():
+                    if k != "Normal":
+                        if v == 1:
+                            self.log(f"Unit State Error: {v}", LogLevel.CRITICAL)
+                    else:
+                        if v == 0:
+                            self.log("Unit State Error: Normal Operation Error Flag", LogLevel.CRITICAL)
```
sbasu107, line 284: Explicitly parse the unit_status dictionary and handle only flag bits while ignoring reserved bits.

subsystem/interlocks/interlocks.py
```diff
@@ -232,8 +271,29 @@ def update_data(self):
...
+                self._check_terminal_status(
+                    input_terms,
+                    self.driver.IN_STATUS,
+                    "Input")
+                
+                # check output terms
+                self._check_terminal_status(
+                    output_terms,
+                    self.driver.OUT_STATUS,
+                    "Output")
```
sbasu107, line 296: call the new `_check_terminal_status` method using `self.driver.IN_STATUS` and `self.driver.OUT_STATUS` for input and output terminals, respectively

instrumentctl/G9SP_interlock/g9_driver.py
```diff
@@ -79,42 +78,31 @@ def setup_serial(self, port, baudrate=9600, timeout=0.5):
...
-                self.log(f"Serial connection established on {port}", LogLevel.INFO)
-            except serial.SerialException as e:
+                    )   
+            except serial.SerialException:
                 self._close_serial()
-                self.log(f"Failed to open serial port {port}: {str(e)}", LogLevel.ERROR)
         else:
             self._close_serial()
-            self.log("No port specified", LogLevel.WARNING)
+            raise ConnectionError
```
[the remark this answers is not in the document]
sbasu107: Fixed

subsystem/interlocks/interlocks.py
```diff
@@ -232,8 +273,39 @@ def update_data(self):
...
+
+                # Check if unit status contains an error and logs message
+                if '__error__' in unit_status:
+                    self._set_all_indicators('red')
+                    self.log(unit_status['__error__'], LogLevel.CRITICAL)
+                    if current_time - self.last_error_time > (self.update_interval / 1000):
+                        self.last_error_time = current_time
+                        self._adjust_update_interval(success=False)
+                        self.parent.after(self.update_interval, self.update_data)
+                        return
```
ColeMov, line 287: Checks for error flag in unit status and logs corresponding message.

PR #67, opened 2025-06-02
Feature/refresh logs

arundhatisingh17:
I adjusted the logic for creating a new log file every 8 hours to avoid overloading the WebMonitor's server with an extensively long log file. I tested the implementation by temporarily reducing the time interval to one minute, and it correctly created a new log file at each interval.


utils.py
```diff
@@ -54,12 +61,17 @@ def log(self, msg, level=LogLevel.INFO):
...
 
             # write to log flie if enabled
-            if self.log_to_file and self.log_file:
-                try:
-                    self.log_file.write(formatted_message)
-                    self.log_file.flush()
-                except Exception as e:
-                    print(f"Error writing to log file: {str(e)}")
+            if self.log_to_file:
+                now = now = datetime.datetime.now()
```
bwalkerMIR, line 65: Why do you have now=now?

dashboard.py
```diff
@@ -103,7 +103,7 @@ def setup_main_pane(self):
         self.root.grid_rowconfigure(0, weight=1)
         self.rows = [tk.PanedWindow(self.main_pane, orient='horizontal', sashrelief=tk.RAISED) for _ in range(5)]
         for row_pane in self.rows:
-            self.main_pane.add(row_pane, stretch='always')
+            self.main_pane.add(row_pane, stretch="always")
```
bwalkerMIR, line 106: Since the whole file uses single quotes this is an unecessary change unless we have a good reason for changing this

utils.py
```diff
@@ -54,12 +61,17 @@ def log(self, msg, level=LogLevel.INFO):
...
+                if self.log_start_time == None or (now - self.log_start_time).total_seconds() > 8*60*60:
```
bwalkerMIR, line 66: Why do you need to call self.setup_log_file() when log_start_time == none? This strikes me as odd because the code still ran fine before when there was no start time, so I'm surprised you would need to call it here.

bwalkerMIR (approved):
This PR67 looks good and Arundhati has tested it on her machine. I am hoping this will prevent the website from crashing this weekend during our pump down. This should result in a new file created every 8 hours.

PR #68, opened 2025-06-13
Feature/interlocks logging

arundhatisingh17:
We need to log the input terminal data flags along with the output terminal data flags to determine the status of all the systems on the interlock panel in the WebMonitor.


bwalkerMIR (approved):
This is a simple change that allows the web monitor to read from the dashboard. Changes look good.

PR #70, opened 2025-06-26
Feature/add feature for logs

arundhatisingh17:
Created a feature to select a separate log level for the saved log file. The saved log file offers the user to select from "VERBOSE" and "DEBUG" with the default set to DEBUG.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -65,6 +65,8 @@ class Logger:
             if self.log_start_time == None or (now - self.log_start_time).total_seconds() > 8*60*60:
                 self.setup_log_file()
             try:
+                if not self.log_file:
+                    return  # No log file available
                 file_formatted_message = f"[{timestamp}] - {level.name}: {msg}\n"
                 self.log_file.write(file_formatted_message)
                 self.log_file.flush()


dashboard.py
```diff
@@ -264,10 +265,36 @@ def create_log_level_dropdown(self, parent_frame):
         log_level_dropdown.set(current_level.name) 
         log_level_dropdown.bind("<<ComboboxSelected>>", self.on_log_level_change)
```
arundhatisingh17, line 267: Added a “File Log Level” dropdown to `parent_frame` with options DEBUG and VERBOSE. When the user picks a level, "on_file_log_level_change" will be called to update self.logger.file_log_level accordingly.

utils.py
```diff
@@ -48,30 +49,28 @@ def setup_log_file(self):
...
         """ Log a message to the text widget and optionally to local file """
         if level >= self.log_level:
             timestamp = datetime.datetime.now().strftime("%H:%M:%S")
             formatted_message = f"[{timestamp}] - {level.name}: {msg}\n"
-            
+        
             # Write to text widget
             self.text_widget.insert(tk.END, formatted_message, ("log",))
             self.text_widget.tag_config("log", font=("Helvetica", 9))  # Set font size
             self.text_widget.see(tk.END)
```
arundhatisingh17, line 62: If file logging is enabled and the message’s level is above or equal to the file_log_level threshold, write it to the saved log file.

utils.py
```diff
@@ -152,7 +151,7 @@ def __init__(self, parent, width=300, height=200):
         )
 
         self.file_logging_enabled = True
```
arundhatisingh17, line 153: Initialized the frame’s Logger with the thresholds for the display and saved log file set to DEBUG by default.

bwalkerMIR:
@codex review

PR #72, opened 2025-07-09
Feature/create global dict

arundhatisingh17:


bwalkerMIR:
There are unnecessary duplicate module imports on lines 161 through 165 in utils.py:

image

bwalkerMIR:
Minor point, but it would probably be better to avoid trailing spaces in the keys for the dictionary or it's going to be prone to errors and mistyping:

**Dictionary keys with trailing spaces**
- Keys such as "Cathode A - Heater Current: " include trailing spaces.
- A small mismatch in strings will raise a KeyError, making the API fragile and harder to use.

bwalkerMIR:
I also think that the webmonitor dictionary doesn't log the CCS temperatures yet. Is there a reason it omits those temperatures?
image

utils.py
```diff
@@ -48,30 +65,72 @@ def setup_log_file(self):
...
+        """Setup a new log file and web monitor log file in the 'EBEAM_dashboard/EBEAM-Dashboard-Logs/' directory."""
+        try:
+            # Use the EBEAM_dashboard directory
+            base_path = os.path.abspath(os.path.join(os.path.expanduser("~"), "EBEAM_dashboard"))
+            log_dir = os.path.join(base_path, "EBEAM-Dashboard-Logs")
+            os.makedirs(log_dir, exist_ok=True)
+            
+            # Create the web monitor log file with the old naming pattern
+            webMonitor_log_file_name = f"webMonitor_log_{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}.txt"
+            # close the existing log file at intervals of 8 hours and create a new one
```
bwalkerMIR, line 79: I think this comment is incorrect. I believe your web monitor log file is closed at intervals of 4 hours, not 8 hours.

arundhatisingh17:
> **Logger required assumption: Several subsystems still call self.logger.update_field unconditionally**. Instantiating these subsystems without a logger raises exceptions
> 
> G9Driver checks for a logger before updating fields. The logger respects the file‑logging toggle and only writes web‑monitor logs when file logging is enabled. but ProcessMonitorSubsystem, CathodeHeatingSubsystem, and VTRXSubsystem still call self.logger.update_field unconditionally. If instantiated without a logger, they will raise an AttributeError.


I have added a check for all the subsystems.

arundhatisingh17:
> **Log-level handling: Logger.log writes all messages to file regardless of log_level**, which may be intentional but differs from GUI filtering behavior.
> 
> @arundhatisingh17 can you remind me if we intentionally decided to do this? So the log level only control what's displayed but not on the log?


Yes, this was intentional. I remember talking about being able to independently control the log level displayed on web monitor/dashboard from the web monitor and dashboard separately. We must log all lines on the file to be able to do so.

bwalkerMIR:
@codex review

bwalkerMIR:
@codex review

bwalkerMIR:
@codex review

bwalkerMIR:
A general comment that I'm not going to address now, but several of the method names created for the dictionary logger are very vague and should have been more specific to make the code more readable. For example dashboard.py now has CLEAR_MAP, which is only for the dictionary logger but makes no indication of that in the name. Same with the logger functions of logger.clear_value and update_field.

bwalkerMIR (approved):
Approving PR 72 after merging in develop and making some edits. A future PR should adjust the dict logger methods to have clearer more specific names, per python convention of code being highly readable.

PR #73, opened 2025-07-09
Fix for initial spike in vacuum graph

ColeMov:
Initial spike on vacuum graph that would show up on dashboard startup was removed. Also, we will now use raw pressure sensor data for the pressure value if it is valid to avoid rounding.
IMG_3335

### DIFF
--- a/main.py
+++ b/main.py
@@ -185,7 +185,7 @@ def config_com_ports(saved_com_ports):
     # Close the PyInstaller splash if running as bundled executable
     if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):
         try:
-            import pyi_splash # type: ignore[import]
+            import pyi_splash
             pyi_splash.close()
         except ImportError:
             pass


subsystem/vtrx/vtrx.py
```diff
@@ -65,13 +65,7 @@ def __init__(self, parent, serial_port='COM7', baud_rate=9600, logger=None):
         self.stop_event = threading.Event()
         self.last_data_received_time = time.time()
         self.last_gui_update_time = time.time()
-        
-        current_time = datetime.datetime.now()
-        self.full_history_x = [current_time]
-        self.full_history_y = [1e3]
-        self.x_data = [current_time]
-        self.y_data = [1e3]
```
ColeMov, line 73: Initial fake first value that was creating initial spike.

subsystem/vtrx/vtrx.py
```diff
@@ -240,6 +234,8 @@ def handle_serial_data(self, data):
         try:
             pressure_value = float(data_parts[0])   # numerical pressure value
```
bwalkerMIR, line 235: Shouldn't we get rid of this line since you reassign pressure_value a few lines below this?

subsystem/vtrx/vtrx.py
```diff
@@ -451,7 +447,8 @@ def update_gui(self, pressure_value, pressure_raw, switch_states):
             for idx, state in enumerate(switch_states):
                 canvas, oval_id = self.circle_indicators[idx]
                 canvas.itemconfig(oval_id, fill='#00FF24' if state == 1 else 'grey')
-            
+            subsystem_bits = ''.join(str(bit) for bit in switch_states)
+            self.log(f"VTRX States: {subsystem_bits}", LogLevel.DEBUG)
```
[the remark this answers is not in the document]
bwalkerMIR: If develop already has this in it, then it shouldn't show up here in files changed. Did you pull the latest develop into this branch?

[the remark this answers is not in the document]
ColeMov: I had made this branch prior to this change being merged into develop. I'm not sure if that would make it show up in the file change but I can double check when I'm in today.

subsystem/vtrx/vtrx.py
ColeMov: Yes, good point. This was leftover from the original implementation.

subsystem/vtrx/vtrx.py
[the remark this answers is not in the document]
ColeMov: Those two lines are functionally the same as in develop. I manually added them in case they would have been removed when merged which is why I think they are marked as a different.

subsystem/vtrx/vtrx.py
```diff
@@ -238,8 +232,9 @@ def handle_serial_data(self, data):
             return
         
         try:
-            pressure_value = float(data_parts[0])   # numerical pressure value
```
ColeMov, line 241: Removed redundant code.

bwalkerMIR:
@codex review

PR #77, opened 2025-09-29
Bugfix/dummy port console error

MathomJohnson:
This PR fixes this issue of errors being logged to the console when running dashboard with dummy ports. This issue was with COM5 (TempControllers) and COM7 (PMON) because they use a different connection method than the other subsystems. PMON and TempControllers use the ModbusSerialClient which catches exceptions internally and prints them out to sys.stderr. I solved this issue by re-routing messages to sys.stderr to dashboard log only within the "connect" method which is where the errors were coming from.

I tested my solution by first, checking that the errors no longer populated the console when running the dashboard. Second, exported the dashboard log and checked that the errors were actually being re-routed there.

Here is the link to the Asana task to learn more about the issue: https://app.asana.com/1/492353873344076/project/1211223975008499/task/1209968122177038?focus=true

### DIFF
--- a/instrumentctl/DP16_process_monitor/DP16_process_monitor.py
+++ b/instrumentctl/DP16_process_monitor/DP16_process_monitor.py
@@ -8,6 +8,17 @@ from pymodbus.exceptions import ModbusIOException
 from typing import Dict
 
 class DP16ProcessMonitor:
+    class StderrRedirector:
+        def __init__(self, logger, level=LogLevel.ERROR):
+            self.logger = logger
+            self.level = level
+        def write(self, msg):
+            msg = msg.strip()
+            if msg:
+                if self.logger:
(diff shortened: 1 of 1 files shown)


instrumentctl/DP16_process_monitor/DP16_process_monitor.py
```diff
@@ -66,49 +78,49 @@ def connect(self):
...
-                working_units = set()
-                for unit in self.unit_numbers:
-                    status = self.client.read_holding_registers(
-                        address=self.STATUS_REG,
-                        count=1,
-                        slave=unit
-                    )
-                    if not status.isError():
-                        working_units.add(unit)
+        import sys
```
bwalkerMIR: imports should always be at the top of the file

PR #87, opened 2026-03-06
Skip DP16 driver if dummy port detected

bwalkerMIR:
This is an alternative attempt to get rid of the dummy com port errors on the terminal when running the dashboard with dummy ports instead of real serial ports. 

A separate PR https://github.com/uw-loci/EBEAM_dashboard/pull/77 Bugfix/dummy port console error attempts to fix this error by rerouting stderr during connection attempts, but that would reroute any writes to stderr to the log file in an unpredictable way depending on the error status of PMON. 

This new PR instead has the PMON subsystem first detect if there is a dummy com port (or no port) and skips instantiating a DP16 driver object if there is no real serial device. The driver file is untouched. This does not affect reconnect attempts if com was later lost with a real serial port because the DP16 driver would have already been recreated and will try to reconnect. This only deals with whether the DP16 driver object is created in the first place when there is only a dummy or no port.

### DIFF
--- a/citestfiles/ProcessMonitor/process_monitor_test.py
+++ b/citestfiles/ProcessMonitor/process_monitor_test.py
@@ -1,0 +1,138 @@
+import enum
+import importlib.util
+import os
+import sys
+import types
+import unittest
+from unittest.mock import MagicMock, patch
+
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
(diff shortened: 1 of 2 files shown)


bwalkerMIR:
Closing this PR because it should be tested more before merging since Tianrui thought he saw the dashboard hang under some circumstance, but our priority is not testing dummy ports at the moment so closing. Likely will reopen later.

PR #98, opened 2026-03-19
CCS: Guard OVP against missing response in 9104 power supply driver

### LINKED ISSUE #94: CCS: Guard OVP response before checking for OK
**<sub><sub>P1 Badge</sub></sub>  Guard OVP response before checking for OK**

If `send_command("GOVP")` returns `None` (e.g., supply disconnected, serial timeout), `if "OK" in response` throws a `TypeError` instead of returning `None`. That exception propagates through `validate_voltage()` callers like `set_voltage()`/`set_output()`, so control actions fail with an unexpected crash path precisely when the device is unavailable.

Useful? React with 👍 / 👎.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/uw-loci/EBEAM_dashboard/pull/93#discussion_r2937020397_

ColeMov:
Addresses issue #94 

P1 Badge Guard OVP response before checking for OK

If send_command("GOVP") returns None (e.g., supply disconnected, serial timeout), if "OK" in response throws a TypeError instead of returning None. That exception propagates through validate_voltage() callers like set_voltage()/set_output(), so control actions fail with an unexpected crash path precisely when the device is unavailable.

### DIFF
--- a/instrumentctl/power_supply_9104/power_supply_9104.py
+++ b/instrumentctl/power_supply_9104/power_supply_9104.py
@@ -557,18 +557,21 @@ class PowerSupply9104:
         command = "GOVP"
         response = self.send_command(command)
 
-        for attempt in range(self.MAX_RETRIES):
-            if "OK" in response:
-                try:
-                    # split the response and take the part before 'OK'
-                    ovp_str = response.split('\r')[0]
-                    # convert to integer, then to a float
-                    ovp_volts = int(ovp_str) / 100.0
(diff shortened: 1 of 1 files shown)


bwalkerMIR:
@codex review

bwalkerMIR (approved):
This change looks good. Fixes issue if none or empty comes in as response. Also gets rid of a pointless for loop. Cole tested on hardware with 2 systems running.

PR #100, opened 2026-03-19
Bugfix/venv

bwalkerMIR:
This branch includes small but important change to the readme clarifying that python 3.11 must be used currently due to some dependency requirements, and that .venv should be created (not venv withouth the dot). This also adds .venv to the .gitignore as well.


bwalkerMIR (commented):
Minimal changes to readme and .gitignore only so shouldn't change how dashboard runs.

PR #106, opened 2026-04-18
Update Test Environment to Use Windows and Python 3.11.9

sbasu107:
This PR is a test to see if we can use a windows environment and Python 3.11.9 for the automated tests, instead of Ubuntu, and python 3.12.4, which we are not using. On this repository, we need to make a pull request in order to trigger automated tests for a particular branch.

GIthub Actions supports WIndows Server Edition, which for Server 2025, uses the same kernel as the laptop windows version (WIndows 11 Education). This works for the Python Unit Tests we are running in out Actions Environment.

I also added a notice to the README that the Python Foundation will end security support of Python 3.11 on Ocober 2027. This will be important for whoever is working on this code at that time.

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v3
       with:
-        python-version: '3.11' 
+        python-version: '3.11.9' 
         
     - name: Install dependencies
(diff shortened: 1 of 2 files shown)


bwalkerMIR:
@codex review

bwalkerMIR (approved):
This is a small focused PR and looks good. Approved.

PR #120, opened 2026-05-04
Bugfix: PMON Thread Leak

danielegelhoff13:
<h2>Overview</h2>
Refactors the logging that is done by the daemon thread responsible for polling the PMON CI. Previously, this logging was done from the background thread, which is not best practice with Tkinter. Logging now uses a queue that the background thread pushes logs to, and the main thread flushes this queue every time the Dashboard updates its values.

<h3>New Behavior</h3>

No TclErrors from the "poll_all_units" thread are printed to the console on Dashboard shutdown. Additionally, since the daemon thread closes cleanly, this PR mitigates concerns with the thread holding on to resources, such as COM port.

<h3>Tests Run</h3>

TODO: run a full hardware test, results will be here: https://docs.google.com/document/d/1HeVK_dALBxRzzOwoVtnQJUaDIUz10hf4UFXZV0RhCP8/edit?tab=t.0

### DIFF
--- a/instrumentctl/DP16_process_monitor/DP16_process_monitor.py
+++ b/instrumentctl/DP16_process_monitor/DP16_process_monitor.py
@@ -53,7 +53,7 @@ class DP16ProcessMonitor:
         self.last_good_readings = {unit: None for unit in self.unit_numbers}
         self.consecutive_connection_errors = 0
 
-        self._is_running = True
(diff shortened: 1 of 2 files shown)


danielegelhoff13:
Closed this PR after @chott24 overhauled the PMON driver to use pyserial instead of pymodbus.

PR #130, opened 2026-05-20
Bugfix: Supabase Failure Safguards

### LINKED ISSUE #97: Throttle Supabase retries after failed insert
**<sub><sub>P1 Badge</sub></sub>  Throttle Supabase retries after failed insert**

In `Logger.log_dict_update`, the 2-second rate limit is only applied after a successful insert because `last_supabase_write` is updated only in the `write_ok` branch. If Supabase is reachable at startup but starts returning errors (or times out), every status update immediately retries instead of waiting 2 seconds, which can flood logs with warnings and block the UI loop with repeated network calls under outage conditions.

Useful? React with 👍 / 👎.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/uw-loci/EBEAM_dashboard/pull/84#discussion_r2949422463_

### LINKED ISSUE #102: If supabase is missing (or broken), importing utils will fail and potentially crash app at startup
Severity: high
File / function: utils.py module import path (from db.supabase_client import SupabaseClient), db/supabase_client.py module import
Problem: Supabase is now imported at module load time, not lazily. If supabase is missing (or broken), importing utils will fail before any try/except in Logger.__init__ can handle it. This can crash the app at startup. 
Why it matters: This turns an optional logging integration into a hard runtime dependency and creates a startup regression risk.
Concrete fix: Move the Supabase import inside Logger.__init__ (or a helper), catch ImportError, and degrade gracefully (self.supabase_client = None) while emitting one warning.

### LINKED ISSUE #103: If Supabase/network is slow can freeze UI potentially
Severity: high
File/function: utils.py → Logger.log_dict_update
Problem: Supabase writes are synchronous in the hot path (insert_status_log is called inline from log_dict_update), with no timeout or async isolation. If Supabase/network is slow, this blocks the caller thread. In this app that can freeze UI/control responsiveness because logging is called during status updates.
Why it matters: A telemetry outage can degrade core app behavior, not just remote logging.
Concrete fix: Move Supabase writes to a background worker/queue (non-blocking enqueue from log_dict_update), and apply a strict network timeout + bounded retry in the worker.

### LINKED ISSUE #122: Supabase "connected" output printed twice.
Two separate Logger instances are created when the Dashboard starts: one for the COM ports configuration stage, and one for the main Dashboard. In its constructor, the Logger prints the Supabase output, which is why we are getting repeated prints.

A simple fix would be to reuse the Logger instance from the COM port configuration stage.

chott24:
Addresses a number of small supabase issues with the dashboard on initialization / load if supabase fails. 

Codex Overview of initial issues and changes:
Supabase was a hard startup dependency
Previously, utils.py (line 32) imported SupabaseClient at module load time. If supabase or one of its dependencies was missing or broken, importing utils could crash the dashboard before Logger had a chance to catch anything.
Fix: moved Supabase creation behind _create_supabase_client(), called inside Logger.__init__, with graceful fallback to supabase_client = None.

Supabase writes blocked the status-update path
Logger.log_dict_update() used to call insert_status_log() inline. A slow network or Supabase outage could block the UI/control thread.
Fix: replaced inline writes with a background worker in utils.py (line 248). The current design keeps only the latest pending status snapshot, which is lighter than a full queue and fits telemetry semantics better.

No strict Supabase network timeout
The Supabase client used SDK defaults, which could allow remote logging to hang longer than desired.
Fix: db/supabase_client.py (line 31) now creates the client with ClientOptions(postgrest_client_timeout=3.0).

Remote logging could still fail noisily
Worker startup failure, insert exceptions, or failed writes could have escaped or created noisy behavior.
Fix: worker start failures disable Supabase logging without raising into dashboard code, and writes retry once before giving up.

Background thread could write into Tk redirected stdout
After the messages pane attaches, sys.stdout is redirected to a Tk text widget. Background-thread print() calls can be unsafe there.
Fix: Supabase worker/client diagnostics now use _safe_console_write() instead of normal print() in worker-facing paths.

Shutdown did not close the logger lifecycle cleanly
main.py did not consistently close the logger when the Tk loop exited. With a background worker, that matters more.
Fix: main.py (line 197) now closes the logger in a finally block after root.mainloop() and also during quit handling.

Cleanup error path referenced a missing logger method
main.py called logger.exception(...), but this custom Logger class does not implement exception.
Fix: changed that fallback to logger.error(...).

Late callbacks could restart logging after shutdown
Scheduled status updates might still call into Logger during or after shutdown.
Fix: Logger.close() now marks the logger closed, disables file logging, stops pending Supabase work, and prevents late log() or log_dict_update() calls from reopening work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,3 +241,9 @@ Assign reviewers to your PR. Merge.
 ```
 python -m PyInstaller EBEAM_DASHBOARD.spec
 ```
+
+### 10. Connectivity
+If any USB ports are unplugged from the computer or I/O expander at any point, you will have to restart the computer before running the dashboard.
+This is important when you are trying to figure out which COM port any given USB corresponds to. 
+This issue has been observed with PMON and Knob Box, but likely affects more CIs.
+
--- a/citestfiles/G9Driver/g9driver_test.py
+++ b/citestfiles/G9Driver/g9driver_test.py
@@ -55,7 +55,7 @@ class TestG9Driver(unittest.TestCase):
         self.driver.ser.read_until.return_value = msg_with_checksum
         
         try:
-            sitsf, sitdf, g9Active, _ , _, _, _ = self.driver._process_response(msg_with_checksum)
+            sitsf, sitdf, g9Output, _ , _, _, _ = self.driver._process_response(msg_with_checksum)
             self.assertEqual(len(sitsf), self.driver.NUMIN)
             self.assertEqual(len(sitdf), self.driver.NUMIN)
             self.assertTrue(all(bit == 1 for bit in sitsf))
--- a/citestfiles/supabase/supabase_integration_test.py
+++ b/citestfiles/supabase/supabase_integration_test.py
@@ -35,6 +35,7 @@ EXPECTED_DICT_LOGGER_KEYS = {
     "clamp_temperature_A",
     "clamp_temperature_B",
     "clamp_temperature_C",
+    "beam_energy",
 }
 
 
--- a/citestfiles/supabase/supabase_integration_test.py
+++ b/citestfiles/supabase/supabase_integration_test.py
@@ -4,7 +4,9 @@ import io
 import json
 import os
 import shutil
+import subprocess
 import sys
+import threading
 import time
 import unittest
 import uuid
@@ -80,8 +82,20 @@ class TestSupabaseClientUnit(unittest.TestCase):
              patch("db.supabase_client.create_client", return_value=mock_client_instance) as mock_create:
             client = SupabaseClient()
 
-        mock_create.assert_called_once_with(fake_url, fake_key)
+        mock_create.assert_called_once()
+        self.assertEqual(mock_create.call_args.args, (fake_url, fake_key))
+        self.assertEqual(mock_create.call_args.kwargs["options"].postgrest_client_timeout, 3.0)
         self.assertIs(client.client, mock_client_instance)
+
+    def test_init_uses_configured_postgrest_timeout(self):
+        mock_client_instance = MagicMock()
+
+        with patch("db.supabase_client.load_dotenv"), \
+             patch("db.supabase_client.os.getenv", side_effect=self._make_getenv("url", "key")), \
+             patch("db.supabase_client.create_client", return_value=mock_client_instance) as mock_create:
+            SupabaseClient(timeout_seconds=1.25)
+
+        self.assertEqual(mock_create.call_args.kwargs["options"].postgrest_client_timeout, 1.25)
 
     def test_insert_status_log_returns_true_on_success(self):
         mock_client_instance = MagicMock()
@@ -108,12 +122,11 @@ class TestSupabaseClientUnit(unittest.TestCase):
              patch("db.supabase_client.create_client", return_value=mock_client_instance):
             sb = SupabaseClient()
 
-        with patch("builtins.print") as mock_print:
+        with patch("db.supabase_client._safe_console_write") as mock_write:
             result = sb.insert_status_log({})
 
         self.assertFalse(result)
-        printed = " ".join(str(c) for c in mock_print.call_args_list)
-        self.assertIn("Supabase insert error", printed)
+        self.assertIn("Supabase insert error", mock_write.call_args[0][0])
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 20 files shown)


chott24:
Addresses Issues:
Throttle Supabase retries after failed insert #97 
If supabase is missing (or broken), importing utils will fail and potentially crash app at startup #102 
If Supabase/network is slow can freeze UI potentially #103

chott24:
Latest commit addressed Issue #122 Supabase "connected" output printed twice

chott24:
@codex review

chott24:
The dashboard was tested with the following configurations:

Valid launch on lab laptop - Success
No .env file - Success
Invalid env SUPABASE_API_URL - Success
Invalid env SUPABASE_API_KEY - Success
Ethernet and wifi disconnected from lab laptop - Success

All issues were handled properly by the dashboard. It did not crash under any of these tests, and logged issues correctly as far as I could tell.

chott24:
This PR is ready for review @bwalkerMIR

chott24:
Here is a summary of my conversation with codex and what we decided to change within Utils.py


**Why Utils Changed**

The `utils.py` changes were all centered on making Supabase logging optional, non-blocking, and safe during startup/shutdown. Before, Supabase could crash dashboard startup if the package/config was broken, and runtime status updates could block while waiting on network calls.

**Lazy Supabase Setup**

At the top of [utils.py](</c:/Users/cott/Documents/Repos/EBEAM_dashboard/utils.py:32>), Supabase creation was moved into `_create_supabase_client()`:

```python
def _create_supabase_client():
    from db.supabase_client import SupabaseClient
    return SupabaseClient(timeout_seconds=SUPABASE_WRITE_TIMEOUT_SECONDS)
```

This means importing `utils.py` no longer imports Supabase immediately. If Supabase is missing, broken, or misconfigured, `Logger.__init__` catches it and sets `self.supabase_client = None` instead of crashing the dashboard.

**Safe Console Output**

`_safe_console_write()` was added at [utils.py](</c:/Users/cott/Documents/Repos/EBEAM_dashboard/utils.py:20>). This writes to `sys.__stdout__` instead of normal `print()`.

That matters because the dashboard later redirects `sys.stdout` into a Tk text widget. Background threads must not write into Tk widgets directly, so Supabase worker messages bypass redirected stdout.

**Non-Blocking Supabase Logging**

`log_dict_update()` now calls `_enqueue_supabase_update()` instead of writing to Supabase inline: [utils.py](</c:/Users/cott/Documents/Repos/EBEAM_dashboard/utils.py:206>).

The new behavior is:

1. Local/web monitor logging still happens in `log_dict_update`.
2. Supabase logging is only scheduled, not executed there.
3. The actual network write happens in a background worker.

This prevents slow Supabase/network calls from freezing UI or control updates.

**Latest-Status Worker**

The lightweight worker uses one pending status snapshot:

- `_supabase_pending_update`
- `_supabase_event`
- `_supabase_pending_lock`

These are initialized in `Logger.__init__` around [utils.py](</c:/Users/cott/Documents/Repos/EBEAM_dashboard/utils.py:66>).

When a Supabase update is due, `_enqueue_supabase_update()` deep-copies the latest status and stores it as the pending update: [utils.py](</c:/Users/cott/Documents/Repos/EBEAM_dashboard/utils.py:248>).

If Supabase is slow and more updates come in, old pending snapshots are overwritten. That is intentional: these are dashboard status snapshots, so the newest state is more useful than sending stale intermediate states.

**Rate Limiting**

The 2-second rate limit is still enforced, but now it is based on when a write is accepted for sending, not only when a write succeeds:

```python
self.last_supabase_write = now
self._supabase_event.set()
```

That prevents outage conditions from retrying on every status update.

**Retry Behavior**

The worker calls `_write_supabase_with_retry()` at [utils.py](</c:/Users/cott/Documents/Repos/EBEAM_dashboard/utils.py:313>). It tries the Supabase insert up to `SUPABASE_WRITE_MAX_ATTEMPTS`, currently 2 attempts, with a short delay between attempts.

If both attempts fail, it logs safely to console and drops that snapshot. The dashboard keeps running.

**Shutdown Safety**

`Logger.close()` now marks the logger closed and stops remote/file work: [utils.py](</c:/Users/cott/Documents/Repos/EBEAM_dashboard/utils.py:364>).

That prevents late scheduled callbacks from reopening logs or restarting Supabase work after shutdown. `_stop_supabase_worker()` wakes the worker and joins briefly, without hanging forever.

In short: `utils.py` now treats Supabase as optional telemetry. If it works, it sends background status updates. If it fails, the dashboard keeps launching, logging locally, updating status, and shutting down normally.

chott24:
Merge Note: when doing a merge on a test branch with all outstanding PRs there was one conflict in dashboard.py with this PR.

dashboard.py on feature/proposed-dashboard-layout:
``` python
def create_messages_frame(self):
        """Create a scrollable frame for displaying system messages and errors."""
        _msg_row, _msg_w, _msg_h = _messages_frame_layout()
        self.messages_frame = MessagesFrame(self.rows[_msg_row], width=_msg_w, height=_msg_h)
        self.logger = self.messages_frame.logger
```
        
dashboard.py on bugfix/supabase-failure-safeguards:
``` python
def create_messages_frame(self):
        """Create a scrollable frame for displaying system messages and errors."""
        self.messages_frame = MessagesFrame(self.rows[3], width = frames_config[-2][2], height = frames_config[-2][3], logger=self.logger)
        self.logger = self.messages_frame.logger
```

Tested, working, merged version:
``` python
def create_messages_frame(self):
        """Create a scrollable frame for displaying system messages and errors."""
        _msg_row, _msg_w, _msg_h = _messages_frame_layout()
        self.messages_frame = MessagesFrame(self.rows[_msg_row], width=_msg_w, height=_msg_h, logger=self.logger)
        self.logger = self.messages_frame.logger
```

bwalkerMIR:
I accepted the combo of incoming develop and this supabase-failure-safeguards @chott24 to match what you said about the merge conflict and to match your end result

bwalkerMIR:
Chase said: Supabase was tested with feature/beam-energy-WM-log merged since the current web monitor has that corresponding branch deployed

bwalkerMIR:
@codex review

bwalkerMIR (approved):
@chase tested this. This includes multiple supabase fixes that could have caused problems. Codex found no more issues. Code looks reasonable. Runs on my machine (still has console errors to terminal but is related to another system). Approved.

PR #137, opened 2026-05-21
Bugfix: Cathode Logger Schema check

### LINKED ISSUE #123: Prevent logger schema corruption from top-level cathode updates
### Description:
`Logger.update_field()` currently accepts any top-level key in `dict_logger`, including `"cathode"`. Unlike the other top-level logger fields, `"cathode"` is a nested subtree used by `Logger.update_cathode_field()` for cathode A/B/C values.

Because `update_field("cathode", value)` is currently allowed, a caller can replace the cathode subtree with `None`, a scalar, or another invalid shape. After that, `update_cathode_field()` assumes `self.dict_logger["cathode"]` is still a mapping and can raise a `TypeError` during membership checks.

This can affect the cathode heating polling loop. `CathodeHeatingSubsystem.update_data()` calls `update_cathode_field()` for clamp-temperature updates outside the local power-supply `try/except`, and the next Tkinter `after()` callback is scheduled only at the end of the method. If the corrupted cathode root causes an exception there, the periodic cathode update loop can stop.

### Suggested fix:
- Reject `field == "cathode"` in `Logger.update_field()`, matching the existing `clear_value("cathode")` protection.
- Direct callers to use `update_cathode_field()` for cathode subfield changes.
- Add a regression test that `update_field("cathode", None)` raises `KeyError`, leaves the cathode subtree intact, and allows a later `update_cathode_field()` call to succeed.

### Impact:
This is not currently triggered by any in-repo caller found during inspection, but it is a real API-level footgun in central logging code. One bad call can corrupt the in-memory status schema and stop cathode heating periodic updates. The fix should be small and low risk.

### Original Codex Find
**<sub><sub>P2 Badge</sub></sub>  Prevent schema corruption when updating cathode root**

`update_cathode_field` assumes `self.dict_logger["cathode"]` is always a mapping, but `update_field` still permits `field == "cathode"` and can overwrite that subtree with a scalar or `None`. After that, the membership check in `update_cathode_field` throws `TypeError`, and `CathodeHeatingSubsystem.update_data()` invokes this path for clamp-temperature updates outside its local try/except, so periodic updates can break once the root key is overwritten. Rejecting top-level `"cathode"` writes in `update_field` (consistent with `clear_value`) or validating the schema before membership checks would avoid this failure mode.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/uw-loci/EBEAM_dashboard/pull/108#discussion_r3170400253_

chott24:
This PR addresses Issue #123 . It has very simple changes to utils:
1. Adds check that prevents field == "cathode" in update_field(). update_cathode_field() should be used instead.
2. Adds check that ensures the "cathode" dict has not been corrupted.
3. Adds a regression test that tests the above two changes.

### DIFF
--- a/citestfiles/supabase/supabase_integration_test.py
+++ b/citestfiles/supabase/supabase_integration_test.py
@@ -460,6 +460,22 @@ class TestDictLoggerFieldManagement(unittest.TestCase):
         self.assertIn("nonexistent_field", str(ctx.exception))
         self.assertIn("is not a valid key", str(ctx.exception))
 
+    def test_update_field_rejects_cathode_key_without_corrupting_schema(self):
+        with self.assertRaises(KeyError) as ctx:
+            self.logger.update_field("cathode", None)
+        self.assertIn("cathode", str(ctx.exception))
+        self.assertIn("update_cathode_field", str(ctx.exception))
+
+        self.assertIsInstance(self.logger.dict_logger["cathode"], dict)
+        self.logger.update_cathode_field("A", "clamp_temperature", 42.0)
+        self.assertEqual(self.logger.dict_logger["cathode"]["A"]["clamp_temperature"], 42.0)
+
+    def test_update_cathode_field_raises_clear_error_for_corrupted_cathode_schema(self):
+        self.logger.dict_logger["cathode"] = None
+        with self.assertRaises(TypeError) as ctx:
+            self.logger.update_cathode_field("A", "clamp_temperature", 42.0)
+        self.assertIn("cathode schema is corrupted", str(ctx.exception))
+
     def test_clear_value_sets_field_to_none(self):
         self.logger.dict_logger["pressure"] = 5.0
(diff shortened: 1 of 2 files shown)


chott24:
@codex review

chott24:
As a quick test I ran the dashboard and things seem to work fine. I'm not completely sure what to do for testing here. Codex did a review of all calls to logger.update_field() and there are currently none that call it with "cathode" as a parameter. The regression test which does call it with "cathode" reports that it gets sent back and does not allow the call, which is the intended behavior of this PR. 

This is more of a preventative change for any future additions to the code where logger.update_field() is used improperly, than it is a bugfix for current issues. 

This PR is ready for review @bwalkerMIR

bwalkerMIR:
@codex review

bwalkerMIR (approved):
Code looks good. Approved.

PR #141, opened 2026-05-21
Bugfix: BCON COM key & placeholder render

### LINKED ISSUE #111: Beam Pulse import fallback renders placeholder in Process Monitor pane
**Intro:**
In [dashboard.py](/c:/Users/bwalker/Code/EBEAM_dashboard/dashboard.py:1045), the fallback path for a missing `BeamPulseSubsystem` looks for legacy frame name `Beam Steering/Pulse`, then falls back to `Process Monitor`. The current layout defines `Beam Pulse`, not `Beam Steering/Pulse`, so the placeholder label would be rendered in the wrong pane.

Severity: low to medium. It only affects the degraded startup path where Beam Pulse cannot be imported, but in that mode it gives users misleading UI: `Process Monitor` gets a Beam Pulse message and `Beam Pulse` stays empty. The code fix is simple and worth doing.

One related note: `subsystem/__init__.py` imports `BeamPulseSubsystem` unguarded, so some Beam Pulse import failures may crash earlier than this fallback. That is separate and broader than the reported badge issue.

**Description:**
When `BeamPulseSubsystem` cannot be imported, `Dashboard.create_subsystems()` uses the fallback container lookup:

`self.frames.get('Beam Steering/Pulse', self.frames['Process Monitor'])`

The current dashboard layout defines a `Beam Pulse` frame, but does not define `Beam Steering/Pulse`. As a result, the missing-subsystem placeholder label is packed into the `Process Monitor` frame instead of the `Beam Pulse` frame.

Expected behavior:
The fallback placeholder should render inside the `Beam Pulse` pane so users can clearly see that the Beam Pulse subsystem is unavailable.

Actual behavior:
The placeholder appears in the `Process Monitor` pane, while the `Beam Pulse` pane remains empty.

Impact:
This only affects the degraded path where Beam Pulse cannot be imported, but it makes the UI misleading during that failure mode. It may cause users to think the Process Monitor panel is broken or related to the Beam Pulse import failure.

Suggested fix:
Update the fallback branch to target `self.frames['Beam Pulse']` or `self.frames.get('Beam Pulse')` directly, matching the current frame configuration. Add a small regression test or startup-path check that forces `BeamPulseSubsystem = None` and verifies the placeholder is attached to the Beam Pulse frame, not Process Monitor.

Related observation:
`subsystem/__init__.py` imports `BeamPulseSubsystem` unguarded, so some import failures may happen before this fallback branch can run. That should be considered separately from this UI-targeting bug.

**<sub><sub>P2 Badge</sub></sub>  Render BeamPulse import fallback in the Beam Pulse frame**

When `BeamPulseSubsystem` cannot be imported, this fallback writes the placeholder label into `Process Monitor` instead of the Beam Pulse pane, because it looks up `'Beam Steering/Pulse'` and then falls back to `self.frames['Process Monitor']`. In that failure mode, users get a misleading Process Monitor panel while the Beam Pulse frame stays empty; the fallback should target the Beam Pulse frame directly.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/uw-loci/EBEAM_dashboard/pull/88#discussion_r3170202598_

### LINKED ISSUE #112: Fix Beam Pulse COM selector to use persisted BeamPulse key
**Description:**

The dashboard's in-app COM-port configuration panel initializes the Beam Pulse selector from the display alias `Beam Pulse`, but the startup configuration flow registers and persists the subsystem under the canonical key `BeamPulse`.

Current behavior:
- `main.py.SUBSYSTEMS` contains `BeamPulse`.
- Saved COM-port config files therefore store the Beam Pulse port under `BeamPulse`.
- `dashboard.py` correctly initializes the actual `BeamPulseSubsystem` from `BeamPulse`, with a fallback to `Beam Pulse`.
- However, the dashboard COM-port selector initializes the visible Beam Pulse dropdown from `self.com_ports.get('Beam Pulse', '')`.
- When Apply is pressed, the dashboard rebuilds `self.com_ports` from the visible selector keys, which can drop the canonical `BeamPulse` entry and replace it with the display alias `Beam Pulse`.

Expected behavior:
- The Beam Pulse COM dropdown should display the already configured persisted port.
- Applying COM-port changes should preserve the canonical `BeamPulse` key.
- Legacy configs that contain `Beam Pulse` should still be accepted as a fallback.

Why this matters:
- Operators can see a blank or incorrect Beam Pulse COM selector even when a port is already configured.
- Pressing Apply can poison the in-memory COM-port config by removing `BeamPulse` or setting the Beam Pulse value to an empty/wrong port.
- Startup initialization is currently protected by a fallback, so this is not likely to break Beam Pulse on initial launch, but the in-dashboard config workflow is misleading and fragile.

Proposed fix:
- Use `BeamPulse` as the canonical selector key in the dashboard COM-port panel.
- Keep the UI label as `Beam Pulse`.
- Initialize from `BeamPulse`, falling back to legacy `Beam Pulse`.
- Apply changes back under `BeamPulse`.
- Add a small regression test covering canonical and legacy alias lookup.

**<sub><sub>P2 Badge</sub></sub>  Initialize Beam Pulse COM selector from the persisted key**

The new COM-port field is initialized from `self.com_ports.get('Beam Pulse', '')`, but this commit registers the subsystem key as `BeamPulse` in `main.py` (`SUBSYSTEMS`). As a result, the Beam Pulse dropdown appears blank even when a port is already configured, and pressing Apply can overwrite the in-memory config with an empty value. Read both aliases (or the canonical key) when initializing this control.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/uw-loci/EBEAM_dashboard/pull/88#discussion_r3170202600_

chott24:
This PR addresses issues #111 and #112. See their descriptions for more details. 

What has changed in this PR:
- BCON no longer has the import safeguards that were put in during early BCON dashboard development
- BCON has consistent COM confik key support for both 'BeamPulse' and 'Beam Pulse' using the first as the main one.

### DIFF
--- a/citestfiles/startup_logging_test.py
+++ b/citestfiles/startup_logging_test.py
@@ -9,7 +9,11 @@ from unittest.mock import MagicMock, patch
 sys.path.append(os.path.dirname(os.path.dirname(__file__)))
 
 from utils import Logger, LogLevel
-from usr.com_port_config import load_com_ports, save_com_ports
+from usr.com_port_config import (
+    get_beam_pulse_com_port,
+    load_com_ports,
+    save_com_ports,
+)
 from usr.panel_config import load_pane_states
 
 TEST_TMP_ROOT = os.path.join(os.path.dirname(__file__), "_tmp")
@@ -124,6 +128,22 @@ class TestComPortConfigLogging(unittest.TestCase):
         self.assertEqual(loaded, {})
         self.assertIn("Error loading COM ports", logger.error.call_args[0][0])
         mock_print.assert_not_called()
+
+    def test_get_beam_pulse_com_port_prefers_canonical_key(self):
+        com_ports = {
+            "BeamPulse": "COM7",
+            "Beam Pulse": "COM8",
+        }
+
+        self.assertEqual(get_beam_pulse_com_port(com_ports), "COM7")
+
+    def test_get_beam_pulse_com_port_accepts_legacy_key(self):
+        com_ports = {"Beam Pulse": "COM8"}
+
+        self.assertEqual(get_beam_pulse_com_port(com_ports), "COM8")
+
+    def test_get_beam_pulse_com_port_defaults_to_blank(self):
+        self.assertEqual(get_beam_pulse_com_port({}), "")
 
 
 class TestPanelConfigLogging(unittest.TestCase):
--- a/citestfiles/startup_logging_test.py
+++ b/citestfiles/startup_logging_test.py
@@ -38,6 +38,11 @@ class FakeTextWidget:
         return "".join(self.messages)
 
 
+class FakeSubsystem:
+    def __init__(self, *args, **kwargs):
+        pass
+
+
 class TestStartupLogger(unittest.TestCase):
     def setUp(self):
         self.tempdir = os.path.join(TEST_TMP_ROOT, f"case_{uuid.uuid4().hex}")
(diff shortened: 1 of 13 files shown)


chott24:
@codex review

chott24:
@bwalkerMIR this PR is ready for review

bwalkerMIR:
@codex review

bwalkerMIR:
@chott24 this PR fixes a couple BeamPulse issues, but I don't think the backward compatibility changes should be kept. This PR changes the main workflow by singling out BeamPulse just in case someone uses and old "Beam Pulse" name but making extra code and rewriting the flow just for a single case like this seems unnecessary, since the Beam Pulse was only recently added anyways to the codebase. I made a new PR to merge into this one getting rid of the extra functionality for old Beam Pulse. What do you think?

chott24:
> @chott24 this PR fixes a couple BeamPulse issues, but I don't think the backward compatibility changes should be kept. This PR changes the main workflow by singling out BeamPulse just in case someone uses and old "Beam Pulse" name but making extra code and rewriting the flow just for a single case like this seems unnecessary, since the Beam Pulse was only recently added anyways to the codebase. I made a new PR to merge into this one getting rid of the extra functionality for old Beam Pulse. What do you think?

I agree. I think when I made this I was still quite unsure what else was using the key. I probably should have done a sweep of that instead of adding blanket backward compatibility. The code changes look good in that small PR you made, I went ahead and merged them into here. 

This PR is ready for re-review @bwalkerMIR

bwalkerMIR:
@codex review

bwalkerMIR (approved):
Looks reasonable. Approving.

PR #149, opened 2026-05-29
Fix cathode E-stop after transient power-supply readback failure

### LINKED ISSUE #148: Beams E-stop can skip cathode supply after transient readback failure
**<sub><sub>P1 Badge</sub></sub>  Do not drop E-stop eligibility on readback failures**

When `GETD` fails to parse or times out, `get_voltage_current_mode()` returns `(None, None, "Err")` while the serial port may still be open, and the new `_mark_power_supply_unavailable()` path clears `power_supply_status`. That status is also the gate used by `turn_off_all_beams()` before it sends `set_output("0")`, which is invoked from the dashboard Beams E-stop path, so a transient readback failure on an energized supply can cause the E-stop/turn-off-all flow to skip the very supply it still has a handle for until a later successful poll restores the flag.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/uw-loci/EBEAM_dashboard/pull/134#discussion_r3325062291_

bwalkerMIR:
## Summary

This PR preserves the cathode E-stop/readback failure fix separately from the CCS disconnect PR.

It updates `turn_off_all_beams()` so the Beams E-stop path attempts to disable every available cathode power-supply handle, even if the latest telemetry poll marked `power_supply_status[index]` false after a transient readback failure. The shutdown path now prefers the existing unconditional `disable_output()` command and falls back to `set_output("0")` only when needed.

## Root Cause

`update_data()` can mark a supply unavailable when `GETD` readback times out or returns invalid voltage/current data. The previous `turn_off_all_beams()` implementation used that same status flag as an eligibility gate before sending the OFF command, so a transient readback failure could cause the dashboard Beams E-stop flow to skip a supply that still had an open driver handle.

## Validation

- `.venv\\Scripts\\python.exe -m pytest citestfiles/CathodeHeating/beams_off_test.py -v`
- `.venv\\Scripts\\python.exe -m pytest citestfiles/ -v`

### DIFF
--- a/citestfiles/CathodeHeating/beams_off_test.py
+++ b/citestfiles/CathodeHeating/beams_off_test.py
@@ -7,12 +7,23 @@ from subsystem.cathode_heating.cathode_heating import CathodeHeatingSubsystem
 from utils import LogLevel
 
 class TestBeamsOff(unittest.TestCase):
+    def make_power_supply(self, disable_result=True, set_result=True, has_disable=True):
+        methods = ['set_output', 'stop_ramp']
+        if has_disable:
+            methods.append('disable_output')
+        ps = MagicMock(spec=methods)
+        ps.set_output.return_value = set_result
+        ps.stop_ramp.return_value = None
+        if has_disable:
+            ps.disable_output.return_value = disable_result
+        return ps
+
     def setUp(self):
         # Bypass __init__ to avoid Tk and image loading
         self.subsys = object.__new__(CathodeHeatingSubsystem)
         # Inject only what turn_off_all_beams uses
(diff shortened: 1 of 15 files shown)


bwalkerMIR:
Once merged this will fix issue #148

bwalkerMIR:
Note that this should have develop merged in after bugfix/CCS-disconnect has been merged into develop, as the diff relative to develop currently shows changes from both this branch and the parent branch bugfix/CCS-disconnect

chott24 (commented):
Looks good. Codex analysis also agrees with the changes made here.

bwalkerMIR:
@codex review

PR #150, opened 2026-05-31
Normalize BeamPulse COM port key

bwalkerMIR:
## Summary

- Removes legacy `"Beam Pulse"` COM-port key compatibility.
- Uses `BeamPulse` as the single canonical saved config key in startup and dashboard COM-port handling.
- Keeps the dashboard label readable while storing selections under `BeamPulse`.
- Removes tests that existed only for the legacy fallback helper.

## Why

The branch was carrying a special compatibility helper for an old key shape that we no longer want to support. This keeps the COM-port config path consistent with `main.py`'s existing `SUBSYSTEMS` key, `BeamPulse`, instead of preserving two spellings.

## Validation

- Passed: focused `BeamPulse` save/load round-trip check.
- Passed: `python -m py_compile usr/com_port_config.py main.py dashboard.py citestfiles/startup_logging_test.py`.
- Passed: `git diff --check`.
- Blocked: `python -m unittest citestfiles.startup_logging_test` fails during import because this Python environment is missing `matplotlib`, imported by `utils.py` before the tests run (`ModuleNotFoundError: No module named 'matplotlib'`).

### DIFF
--- a/citestfiles/startup_logging_test.py
+++ b/citestfiles/startup_logging_test.py
@@ -9,11 +9,7 @@ from unittest.mock import MagicMock, patch
 sys.path.append(os.path.dirname(os.path.dirname(__file__)))
 
 from utils import Logger, LogLevel
-from usr.com_port_config import (
-    get_beam_pulse_com_port,
-    load_com_ports,
-    save_com_ports,
-)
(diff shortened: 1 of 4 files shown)


chott24 (commented):
Looks good

PR #172, opened 2026-06-22
Return none when ERR

bwalkerMIR:
This PR fixes a cathode temperature display fallthrough in `read_temperature()`. When the E5CN temperature controller reports a cached string error such as `"ERROR"`, the GUI now leaves the clamp temperature label at `ERR` and keeps the plot in the error state instead of falling through to the default missing-reading display of `-- C`. The invalid reading still returns `None`, so logging, plotting data, and overtemperature comparisons continue to exclude it from numeric temperature handling.

It also adds focused regression coverage for the two relevant UI states: cached sensor errors display `ERR`, while genuinely missing cached data still displays `-- C`. This preserves the intended distinction between a known sensor/controller error and no available temperature reading, making the dashboard status clearer without changing the broader temperature-control behavior.

### DIFF
--- a/citestfiles/CathodeHeating/read_temperature_test.py
+++ b/citestfiles/CathodeHeating/read_temperature_test.py
@@ -1,0 +1,48 @@
+import os
+import sys
+import unittest
+from unittest.mock import MagicMock
+
+sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
+
+from subsystem.cathode_heating.cathode_heating import CathodeHeatingSubsystem
+
+
+class FakeStringVar:
+    def __init__(self, value=None):
+        self.value = value
+
+    def set(self, value):
+        self.value = value
+
+
+class TestReadTemperature(unittest.TestCase):
+    def setUp(self):
+        self.subsys = object.__new__(CathodeHeatingSubsystem)
+        self.subsys.temperature_controller = MagicMock()
+        self.subsys.temperature_controller.connected = True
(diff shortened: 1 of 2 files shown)


bwalkerMIR:
Title of Issue discovered (not actually on github issues though): E5CN cached sensor errors are displayed as blank temperature readings

Description:
When `E5CNModbus` returns the string sentinel `SENSOR_ERROR` / `"ERROR"` for an invalid temperature reading, `CathodeHeatingSubsystem.read_temperature()` handles the string branch by setting the clamp temperature label to `ERR`, setting the plot color to `ERROR`, and logging the condition.

However, that branch does not return. Execution falls through to the default path at the end of `read_temperature()`, which immediately resets the same temperature label to `-- C`. As a result, a known sensor/error condition is shown in the GUI as a missing or blank temperature reading instead of the explicit `ERR` state.

This is inconsistent with the numeric over-hard-limit path, which sets `ERR`, marks the plot as an error, logs the issue, and returns immediately. It also weakens the intent of surfacing invalid E5CN readings distinctly from absent data.

Expected behavior:
- Cached string sensor errors should leave the clamp temperature display as `ERR`.
- The plot should remain in the error color state.
- The method should continue returning `None` so invalid readings are not treated as numeric temperatures.

Suggested fix:
- Return `None` immediately after handling the `isinstance(temperature, str)` branch in `read_temperature()`.
- Add a regression test that verifies cached `"ERROR"` leaves the label as `ERR` and does not fall through to `-- C`.

chott24 (commented):
looks good

PR #205, opened 2026-07-07
Bugfix: Directly map heater current to beam current within LUT domain

### LINKED ISSUE #202: Heater current to beam mA should be direct mapping
The current experimental look up table does a super unnecessary conversion from heater current to heater voltage and then using voltage to get to beam mA. But the heater voltage data is way lower accuracy and noisy so this unnecessarily pollutes the conversion in going from heater current to beam mA.

ColeMov:
Resolves issue #202 

Updated LUT functionality to directly map heater current or voltage to beam current depending on which variable is predicted to be the controlling value. Implemented after the Richardson-Dushman fallback so predictions above the LUT domain remain intact. 

When reviewing the LUT documentation it was noted that the LUT is explicitly cleaned into a voltage-keyed function taking only the maximum heater current for each heater voltage 'bin'. This could reduce the overall accuracy of the dataset, particularly when predicting using heater current.

https://github.com/uw-loci/EBEAM_dashboard/blob/a0a06e2fc8347c0c6c1d66a86cdc42082057e428/data/lut/README.md?plain=1#L25-L33

Direct mapping from heater current to beam current would still likely be preferable to avoid multiple interpolations for the same prediction, however for improved accuracy it is worth revisiting how we clean our LUT data.

### DIFF
--- a/subsystem/cathode_heating/README.md
+++ b/subsystem/cathode_heating/README.md
@@ -303,6 +303,8 @@ Manual setpoint handlers now actively update predictions before applying output 
 
 Those handlers call either `update_predictions_from_current()` or `update_predictions_from_voltage()` after validation. LUT selector changes call `refresh_predictions()`, which recomputes predictions from the currently requested setpoint when one exists.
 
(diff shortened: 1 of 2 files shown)


ColeMov:
@codex review

ColeMov:
This has been tested on hardware successfully.

Should be merged into develop only after PR #180 feature/pressure-shutoffs

PR #208, opened 2026-07-08
Bugfix: Supabase Write Frequency

chott24:
This PR lowers the Supabase write frequency from 2s to 3s

The web monitor is configured so that the dashboard writes to supabase every 3 second. Currently, the web monitor graphs cap out in datapoints before their desired time interval because the data is more densely uploaded than it expects (every 2s instead of expected 3s). 

Lowering the supabase write frequency would also reduce supabase storage usage. Currently supabase keeps up to 49 hours of short-term high-frequency dashboard logs. Publishing every 2s for 49 hours means 88,200 writes, writes can be 1.5KB to 4KB which is ~130MB to ~350MB. Reducing the frequency down to 3s, means at maximum 58,800 short term logs are stored (~90MB to ~235MB), giving us more breathing room with our 500MB limit


chott24:
@codex review

chott24:
This PR has been tested on the lab laptop. It successfully communicates to the webmonitor on render. I also launched a local deployment on my computer so I could look at logs, and it seems to be seeing a write from the dashboard every 3s instead of every 2s. 

@bwalkerMIR this PR is ready for review

